PR #12, opened 2022-10-03
Adding support for multiple environments and framework updates

skumar09:
- Added support for test execution on multiple sites or environment and updated the framework 
- Following are the npm run commands for executing tests on different environments or application
`test:milo-chrome` Runs all tests on milo site using chrome browser in gui mode
`test:milo-firefox` Runs all tests on mile site using firefox browser in gui mode
`test:milo-safari` Runs all tests on milo site using safari browser in gui mode
`test:bcom-chrome` Runs all tests on bacom site application using chrome browser in gui mode
`test:bcom-firefox` Runs all tests on bacom site application using firefox browser in gui mode
`test:bcom-safari` Runs all tests on bacom site application using safari browser in gui mode

Note : User can add additional tags to above command to choose specific tests.

Also updated following files to support above commands, and update the framework structure
-`selectors.js`
-`parse.js`
-`playwright-config.js`

### DIFF
--- a/selectors/howto.selector.js
+++ b/selectors/howto.selector.js
@@ -1,0 +1,19 @@
+const { expect } = require("@playwright/test")
+
+exports.HowTo = class HowTo {
+
+  constructor(page){
+
+    this.page = page
+    this.howToBlock = page.locator('.how-to')
+    this.howToHeading1 = page.locator('.how-to h1')
+    this.howToHeading2 = page.locator('.how-to h2')
+    this.howToFirstAnswer = page.locator('//ol/li[1]/div')
+    this.howToFourthAnswer = page.locator('//ol/li[4]/div')
+  }
+
+  
+
+}
+
+
--- a/utils/env.js
+++ b/utils/env.js
@@ -1,0 +1,7 @@
+
+exports.Env = class Env {
+
+    baseUrl = process.env.baseUrl
+
+
+}
--- a/utils/globalSetup.js
+++ b/utils/globalSetup.js
@@ -1,0 +1,17 @@
+const FullConfig = require('@playwright/test')
+const dotenv = require('dotenv')
+
+/**
+ * globalSetup for enviroment
+ * 
+ */
+module.exports = async function  globalSetup(config=FullConfig) {
+
+    if(process.env.test_env){
+        dotenv.config({
+            path: `.env.${process.env.test_env}`,
+            override: true
+        })
+    }
+
+}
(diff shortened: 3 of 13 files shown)


utils/env.js
```diff
@@ -0,0 +1,7 @@
+
+exports.Env = class Env {
+
+    baseUrl = process.env.baseUrl
+
+
```
rgclayton, line 6: super nit: remove extra line spaces

utils/globalSetup.js
```diff
@@ -0,0 +1,17 @@
...
+ */
+module.exports = async function  globalSetup(config=FullConfig) {
+
+    if(process.env.test_env){
+        dotenv.config({
+            path: `.env.${process.env.test_env}`,
+            override: true
+        })
+    }
+
```
rgclayton, line 16: Same here.

super nit: remove extra line spaces

selectors/howto.selector.js
```diff
@@ -0,0 +1,19 @@
...
+    this.howToHeading2 = page.locator('.how-to h2')
+    this.howToFirstAnswer = page.locator('//ol/li[1]/div')
+    this.howToFourthAnswer = page.locator('//ol/li[4]/div')
+  }
+
+  
+
+}
+
+
```
rgclayton, line 19: super nit: remove extra line spaces

rgclayton (approved):

amauch-adobe (commented):
if we are planning on using the .env. files for each different environment is it necessary to have the url property in the features files for the different blocks?

Also, is the continued direction for new environments for testing to add more and more lines for the different browsers and sites in the package.json?  Just wondering if that is standard or best practice or if it would be better to have it as a command line parameter for test runner execution?  Just wanting to know what is best, more simple? 

What are the needs for the two new dependencies?  I can see they are used for creating the base environment URL paths for the dotenv dependency, but I was curious why cross-env is needed for executing on a different site domain?

Other than those questions everything else seemed just fine and approved.

skumar09:
> 
@amauch-adobe : 
- I think we can update or remove url property from feature.spec files in future, since currently it is getting used in test naming tag, so i din't touch it.
- For test script commands, using package.json script section is the practice. We can see if we can still shorten it by adding some condition.
- New dependency dotenv and cross_env are basically required for building and executing scripts on multiple environments

auniverseaway:
At first I wasn't a fan of this, and now re-reading, I think this is a good idea, but needs a little refinement.

+100 to the .env concept, but I think we may be able to simplify it (to @amauch-adobe's point).

I'm not in love with all the extra scripts added to package.json. I'd like to tip toe into those needs as we have them. Especially if we can control most of that through CLI parameters.

Maybe @amauch-adobe and you can get on a call with me and we can poke at this? I think it's really close.

PR #13, opened 2022-10-04
Smail

auniverseaway:


auniverseaway:
@skumar09 This should be everything. Retains all your ideas for:

1. Decoupling domains and paths
2. Support for multiple ENVs for a given path.
3. Moved selectors into separate files (i.e. marquee.selectors.js)

This loses a lot of the npm env dependencies and the extra npm scripts.

skumar09 (approved):
Smail :-)

amauch-adobe (commented):
Approved.  Though as a thought with selector files for each block, it might be good to have a selectors file that holds standard selectors used across many tests.  Obviously, right now we don't necessarily have any yet.

PR #15, opened 2022-10-06
Columns Test Added

amauch-adobe:
- Added Columns Test
- Added environments bacomblog and blog to envs.js
- Added workspace settings file to .gitignore for VSCode


amauch-adobe:
These were the comments from the slack conversation in #milo-automation:

Santosh:
 i am still thinking / bit debating on way of using <compoent>-selector.js
/ const row = page.locator(`${selectors[props.tag]} >> .row`).first();

we should go with class object for selectors

Aaron:
So are you thinking that we shouldn’t use the idea of selectors.js and instead create a page object class?
I could make that change, but I think the reason we had the selectors.js was so we could tie it into the tags from the spec file.  Am I thinking incorrectly about that?

Chris M:
I dislike the idea of a class object abstraction for selectors. If you watch that video, even our current way of needing three files to get one test out the door seems a bit... much.
The proposal you had Santosh in your original PR required passing around objects, and there wasn't really any logic being done in that class file.

Santosh:
my thinking was ,  not to put too much code / logic in test files
looking from long run / maintenance perspective, how can we reduce changing at multiple scripts for a change in selector definition...

Chris M. :
Selectors are out of the test file. Adding a class isn't going to make that any better.

Santosh:
i would also think of adding re-usable function there or somewhere , rather in test scripts... 
No class is okie,/  question is where to place common validation functions ?
should we have separate js?

So with that discussion I'm going to adjust the test so that the .row/.col class selectors are actually in the columns-selectors.js file.

amauch-adobe:
@skumar09 , I fixed the github failures for the columns test.  You can see now that the playwright tests pass in GitHub below.  I also moved the selectors you mentioned into the columns.selectors.js file.

auniverseaway (approved):

auniverseaway:
@amauch-adobe good point on the private slack convos. They need to be here. I'll get better at that.

PR #20, opened 2022-10-13
Failed block checking on consumer key pages

amauch-adobe:
Do not merge!

This PR is for reviewers to look over for the first iteration of reviewing.

Ticket: [MWPW-118685](https://jira.corp.adobe.com/browse/MWPW-118685)

- Added failed block checking for a list of key URL pages from consumers
- Failed block checking is signified by the "data-failed=true"

SharePoint file of URLs:  [consumer-keyurls.xlsx](https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7B1624901c-52bf-4a91-9e32-3004fb253532%7D&action=editnew&cid=5f5cf48d-4c71-4d70-ac73-622f18288307)


amauch-adobe:
I added failed.block.spec.js as a possible option for housing URLs and environments.  Will require parse.js file change to allow multiple paths.

amauch-adobe:
I also left the URL for the spreadsheet as a hard-coded value since I felt like it would always be housed there.

amauch-adobe:
updated test to remove the JSON parser as discussed during a short review with @skumar09 and @auniverseaway.  We will instead update parse.js to allow multiple paths.

auniverseaway (changes requested):
1. Please use the new array format for passing in multiple paths.
2. We shouldn't do more than 10 pages per consumer (at least for now). Pages should be meaningfully different. I'd probably argue no more than 5 for consumers that don't have a lot of traffic (Stock).
3. We shouldn't be testing pages we don't own (i.e. AEM pages you have listed for BACOM).

amauch-adobe:
For point 3, I knew we weren't going to test those pages, in my first design of this test I had a conditional that would check if a URL was "available" or not, meaning migrated over yet to Milo/Franklin.  I had that value in the spreadsheet concept that would be JSON fetched.  Anyways, thanks for reminding me that I left those in and needed to remove them.  I'll get on the other two.

amauch-adobe:
I closed this PR since it added ESLint when it should have been in a different branch of fixes.  To keep it clean I opened a new PR with the requested changes from @auniverseaway applied there.  The new PR is [here](https://github.com/adobecom/janus/pull/22).

PR #22, opened 2022-10-15
Added Failed Block Test

amauch-adobe:
Failed Block Checker

Ticket: [MWPW-118685](https://jira.corp.adobe.com/browse/MWPW-118685)

- Added failed block checking for a list of key URL pages from consumers
- Failed block checking is signified by the "data-failed=true"


auniverseaway (approved):

features/failedblock.spec.js
```diff
@@ -0,0 +1,19 @@
...
+    {
+      name: '@failedblockCheck',
+      path: ['/customer-success-stories/', '/customer-success-stories/ben-and-jerrys-case-study',
+        '/customer-success-stories/abb-case-study'],
+      envs: '@bacom',
+      tags: '@failed-block',
+    },
+    {
+      name: '@failedblockCheck',
+      path: ['/pages/artisthub/', '/pages/artisthub/get-started', '/pages/artisthub/learn',
```
auniverseaway, line 13: nit: I'd do carriage returns to make this easier to read.

tests/failedblockcheck.test.js
```diff
@@ -0,0 +1,21 @@
...
+const selectors = require('../selectors/failedblock.selectors.js');
+
+// Parse the feature file into something flat that can be tested separately
+const { name, features } = parse(failedBlock);
+
+test.describe(`${name}`, () => {
+  features.forEach((props) => {
+    test(props.title, async ({ page }) => {
+      await page.goto(props.url);
+      await page.locator('footer').hover(); // Added to give time for failed block JS to load. Without it, test becomes flaky.
```
auniverseaway, line 13: Interesting.

tests/failedblockcheck.test.js
```diff
@@ -0,0 +1,21 @@
...
+      const failedBlocks = await page.$$(selectors[props.tag]);
+      for await (const failedMessage of failedBlocks.map((failed) => failed.getAttribute('data-reason'))) {
+        console.log(`Failed Block Message: "${failedMessage}" : Page URL: ${props.url}`);
+      }
+      expect(failedBlocks).toHaveLength(0);
+    });
+  });
+});
```
auniverseaway, line 21: Do we want to have a failed block in here, too?

amauch-adobe: Wanted to get clarity on your question, do you mean we want to list what failed blocks happened and have that part of the expect() failure messaging?  Or are you asking if we should even have this check in the test?

I list out the failed blocks and their inner HTML messaging in the console in the for loop above, so I didn't think it would make sense to have it in two places.  Though now thinking about it, it would be best for it to be apart of the expect method.   I'll make that change.

features/failedblock.spec.js
amauch-adobe: no worries, good call out.

PR #32, opened 2022-10-31
FaaS Testing

amauch-adobe:
Added FaaS E2E testing, checks up to 5 FaaS forms for each site thus far, in this case, it is only BACOM.

Negative testing and form validation testing was out of scope for the FaaS Milo Block due to all form validation being inherited by the AEM templates from FaaS.  

This testing is essentially validating the FaaS block renders for the desired form templates, form filling works, and submissions are successful in navigating to the desired thank you page.

Addresses this ticket: [MWPW-119374](https://jira.corp.adobe.com/browse/MWPW-119374)


Dli3 (approved):
Lgtm.

PR #33, opened 2022-11-11
Adding modal tests

Dli3:
Adding modal tests.
Ticket: [https://jira.corp.adobe.com/browse/MWPW-119063](url)

The tests will verify the following:

- selecting a modal button will open a modal with a hash in url
- selecting the modal close button closes the modal and hash gets removed in url
- the modal is within viewport when opened 
- the first focus item is the close button when modal opens
- able to close modal with the escape key


tests/modal.test.js
```diff
@@ -0,0 +1,82 @@
...
+
+      /**
+       * There's an issue with webkit where you can't immediately close the
+       * modal with the escape key. The workaround is by tabbing 4 times to
+       * get to the close button, then we can close the modal with the esc key.
+       * This is only an issue in webkit. You can close the modal with the esc
+       * key when the modal opens on Safari.
+       */
+      // eslint-disable-next-line no-underscore-dangle
+      if (page._parent._browserType._initializer.name === 'webkit') {
```
amauch-adobe, line 41: As another option for returning the browser type you can send in the "browser" object at the test method declaration like so:  

test(props.title, async ({ page, browser }) => {

And then you can make a conditional as such:

browser.browserType().name() === 'chromium'

There isn't a right or wrong way to do it, it is just another way.  No need to make a change, just pointing out another option.

tests/modal.test.js
```diff
@@ -0,0 +1,82 @@
...
+        await page.waitForTimeout(1000);
+        let count = 0;
+        while (page.url().includes('#') && count < 10) {
```
amauch-adobe, line 44: Question for clarification:  

Where you said it takes four tabs to get the escape key option to work for webkit to close a modal, why is the loop going 10 iterations?  Is it just an arbitrary number to ensure you eventually get to the other condition of the page not including the '#' mark?  Is there a need for the arbitrary 10, if it is arbitrary?  Why not 6?  Just curious, I'm sure you have reason, I was just wondering.

amauch-adobe (changes requested):
This is awesome, Dennis!  You did a really good job, I had a few comments and potential requested changes.  Please feel free to answer my questions back in the PR.

tests/modal.test.js
Dli3: Thanks for the suggestion. I think that's a better approach. That way we don't need to deal with eslint not liking the underscores.

tests/modal.test.js
Dli3: Even though 4 tab count allowed the esc key to work for webkit majority of the time, I still found it testing at times when testing it. I found that when I used 10 it had a better success rate. 

Having this work around for webkit tests isn't the best approach. I submitted a ticket, [https://jira.corp.adobe.com/browse/MWPW-120941](url) , to add a focus trap. Hopefully this resolves the flaky modal tests on webkit.

amauch-adobe (approved):
Looks great!

PR #40, opened 2022-11-29
GeoRouting E2E Test

amauch-adobe:
Test(s) will cover:

1. As a Swiss visitor, if I visit a non-Swiss page (US) I should be presented with a locale modal.
2. As a Swiss visitor, if I visit a non-Swiss page (DE) I should be presented with a locale modal.
3. If I am a Swiss visitor and I am presented with the locale modal, I should see as many options as there are languages for my country (German, French, Italian).
4. If I click on the page region button (US) an international cookie should be set that corresponds to my region (international = us)
5. If I click outside of the modal, a cookie should not be set.
6. If I navigate to a different page (of the same locale) after clicking outside the modal, I should be prompted to make a decision again.
7. If I navigate to a different page (of the same locale) after choosing a region, I should no longer see the modal.
8. If I navigate to a different region (DE) after choosing a region in the modal (CH_FR), I should see the modal.
9. The detected region options should present the region siblings of the page I am on. I.e. If I am on /products/photoshop I should be presented with /ch_fr/photoshop, etc.
10. If the detected regions do not have the region sibling (ie. 404) the links should point to the root of my site for a given region (/ch_fr/). Context: a given region may not have the same offerings/product as the current page region.
11. If fallback routing is disabled, regional options that do not have a regional page sibling should not be presented to the visitor. Context: Adobe SUMMIT will only be localized in 2-3 languages. We should not offer our out-of-region visitors the ability to "fall back" to their homepage and miss out on Adobe SUMMIT content. They should stay where they are.
12. If GeoRouting is turned off, there should be no modal presented no matter what. A call to geo2 should also not be made.
13. If I am a visitor from Qatar, I should links to the mena_en or mena_ar pages based on the geolocation excel configuration.

Jira Ticket: [MWPW-119375](https://jira.corp.adobe.com/browse/MWPW-119375)


amauch-adobe:
WIP do not merge yet.

skumar09:
@amauch-adobe 

- you are checking the alt-lang value which is good, along with this can you please add one more step for validating exact text message on the geo routing pop-up ( as below)
image
- Also you may want to add one more test script for checking the geo popup for countries where you have more than 2-languages.

amauch-adobe:
@skumar09, thanks for the additional insights, I will do that.

skumar09 (approved):

tests/milo/georouting.test.js
```diff
@@ -0,0 +1,234 @@
...
+import georouting from '../../features/milo/georouting.spec.js';
+import parse from '../../libs/parse.js';
+import selectors from '../../selectors/milo/georouting.selectors.js';
+
+// Parse the feature file into something flat that can be tested separately
+const { name, features } = parse(georouting);
+
+test.describe(`${name}`, () => {
+  features.forEach((props) => {
+    if (props.tag === '@georouting' && props.title.includes('@bacom_live')) {
```
JackySun9, line 26: Why not separate the test into different files

amauch-adobe: The main decision to not separate them was because v2 of the georouting will soon be moved over to BACOM and there won't need to be a separation anymore since the selectors will be the same.

amauch-adobe: It's temporary, I've actually just asked the core Milo team when they will be deploying it to BACOM.

amauch-adobe: @JackySun9, Milo core team came back stating they weren't in a hurry to deploy V2 to BACOM yet since no customers were requesting it.  I have separated out v1 from v2 tests.

JackySun9 (approved):

PR #41, opened 2022-12-07
Nala CI/CD Debugging

amauch-adobe:
- added debug.yml workflow to provide on-demand execution only for HTML reporting results from playwright test executions for specific branches and tags.  The report artifact is uploaded to GitHub actions and can be downloaded to be used against Playwright's Trace Viewer
- added reporter variable to debug.yml workflow 
- added reporter variable setting to run.sh, the execution can use the HTML reporter if desired, otherwise, it will use the test suite's default reporter set in playwright config.

Resolves: [MWPW-122172](https://jira.corp.adobe.com/browse/MWPW-122172)

### DIFF
--- a/.github/workflows/debug.yml
+++ b/.github/workflows/debug.yml
@@ -20,6 +20,7 @@ jobs:
         uses: ./
         env:
           labels: ${{ inputs.tags }}
+          reporter: html
           IMS_EMAIL: ${{ secrets.IMS_EMAIL }}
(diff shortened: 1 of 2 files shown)


amauch-adobe:
For a debugging execution from the branche's changes can be seen here: [Run Nala Debugging #11](https://github.com/adobecom/nala/actions/runs/3635580208)

PR #42, opened 2022-12-13
Added options to run tests against localhost.

Dli3:
Adding options to run Nala tests against localhost 3000 and 6456. 
Ticket: https://jira.corp.adobe.com/browse/MWPW-119683  

In order to test against localhost, the user would have to run the runLocalRegression.sh script. If they don't specify a tag with the -t flag, it'll run all tests. If the 6456 port isn't specified with the -p flag, it'll default to port 3000. 

What the script does: 
1. Updates the envs line in ".feature" files to add either "@localhost:3000" or "@localhost:6456" and comments the existing value out. 
2. It then runs the tests against localhost.
3. It then reverts the envs back to their previous value.

### DIFF
--- a/envs/envs.js
+++ b/envs/envs.js
@@ -6,4 +6,6 @@ module.exports = {
   '@acrobat': 'https://main--acrobat--adobecom.hlx.live',
   '@bacom_prod': 'https://business.adobe.com',
   '@blog': 'https://main--blog--adobe.hlx.live',
+  '@local3000': 'http://localhost:3000',
+  '@local6456': 'http://localhost:6456',
 };
--- a/runLocalRegression.sh
+++ b/runLocalRegression.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+# This script allows users to run regression tests against localhost
+T_FLAG=''
+H_FLAG=''
+B_FLAG=''
+P_FLAG=3000
+
+Help() {
+  echo "This script allows users to run regression tests against localhost.\n"
+  echo "Options:"
+  echo "-t Tag name. Default=all.\n 
+  ie. ./runLocalRegression.sh -t @modal\n"
+  echo "-h Headed mode. Defaults to headless.\n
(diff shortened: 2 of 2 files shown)


Brandon32 (approved):

amauch-adobe (changes requested):
This looks really good!

- One change I would add is the ability to choose a browser to run on.  Meaning, allow the option to choose to run tests locally only on one browser they choose (Chromium, Firefox, or Safari) and then too also keep the option to run on all browsers.  So by default it would run on all browsers headless, but if a specific browser flag is sent in (I would make it similar to how playwright currently allows it, i.e., look [here](https://playwright.dev/docs/test-cli).  
- I think it would also be good to give them the option to run it headless or headed, headless by default
- Also, quick question, was there a reason for choosing to make a separate shell file instead of creating a different conditional or flag within the original run.sh file.  Just thinking that might remove the need to have a separate shell file and instead just have a conditional flag for choosing which flow to execute.

Dli3:
> This looks really good!
> 
> * One change I would add is the ability to choose a browser to run on.  Meaning, allow the option to choose to run tests locally only on one browser they choose (Chromium, Firefox, or Safari) and then too also keep the option to run on all browsers.  So by default it would run on all browsers headless, but if a specific browser flag is sent in (I would make it similar to how playwright currently allows it, i.e., look [here](https://playwright.dev/docs/test-cli).
> * I think it would also be good to give them the option to run it headless or headed, headless by default
> * Also, quick question, was there a reason for choosing to make a separate shell file instead of creating a different conditional or flag within the original run.sh file.  Just thinking that might remove the need to have a separate shell file and instead just have a conditional flag for choosing which flow to execute.

Thanks! I've made the update with the added options.

I created a separate file for clarify and because there are a lot more arguments to handle in the local run script. I thought adding it to the run.sh would get pretty messy having it in nested conditional statements.

amauch-adobe (approved):
All changes look good, thanks.

PR #52, opened 2023-01-18
Fix for modal not showing up

amauch-adobe:
Fix for modal not showing up when directly navigating to the #url when using milolibs url parameter.


Dli3 (approved):
LGTM.

PR #55, opened 2023-01-19
DC converter test

JFernandezAdobe:
Jira ticket: [MWPW-124781](https://jira.corp.adobe.com/browse/MWPW-124781)

Test will cover the E2E conversion flow from PDF to PPT:

1. Upload a test document
2. Wait for conversion to complete
3. Wait for file preview
4. Download the converted file


tests/dc_converter.test.js
```diff
@@ -0,0 +1,45 @@
...
+      await fileInput.setInputFiles(testPDF);
+
+      // Wait for conversion to complete
+      await expect(pdfComplete).toBeVisible({ timeout: 30000 });
+
+      // Wait for file preview
+      await expect(filePreview).toBeVisible();
+
+      // Start waiting for download before clicking. Note no await.
+      const downloadPromise = page.waitForEvent('download');
```
amauch-adobe, line 33: Nice usage of the waitForEvent()!  This is cool!

tests/dc_converter.test.js
```diff
@@ -0,0 +1,45 @@
...
+      const filePreview = page.locator(selectors['@file-preview']);
+      const downloadButton = page.locator(selectors['@download']);
+
+      await expect(fileInput).toBeVisible();
+
+      // Upload a test document
+      await fileInput.setInputFiles(testPDF);
+
+      // Wait for conversion to complete
+      await expect(pdfComplete).toBeVisible({ timeout: 30000 });
```
amauch-adobe, line 27: I'm assuming the increased timeout is because the time for the conversion to complete generally takes longer than the global default of 5 seconds.  Not a requirement, but I sometimes like to provide context for the increase of the timeout in a comment, I have had our higher engineer managers ask about the increased timeouts before and it is good to provide reasons for them.

features/dc_converter.spec.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  name: 'DC Converter',
+  features: [
+    {
+      name: '@frictionless',
+      path: [
+        '/acrobat/online/pdf-to-ppt',
+      ],
+      envs: '@dc_preview',
+      tags: '@converter',
```
amauch-adobe, line 10: Is there a possibility of different types of converters other than just PDF converters?  It might be beneficial to specifically make the tag @pdf-converter or something and then maybe the "name" tag to @converter so it gets tested for all converters.  Using @frictionless is okay, but maybe not necessary since the Frictionless team is over document cloud and the "envs" value already tells us what team or domain it is for.  These are just my thoughts and I could totally be wrong, just an opinion.  How the file currently works just fine.

amauch-adobe (changes requested):
This was awesome! Way to go on the playwright's usage for file uploads and downloads.  I haven't decided just yet, but I'm thinking data files should be housed under a general location, not a breaker for the PR to be merged.  You just happen to be the first person to use a data file, so I'm going to look into what is a general standard with playwright frameworks.  See my other comments though.

envs/envs.js
```diff
@@ -4,7 +4,8 @@ module.exports = {
   '@bacom': 'https://main--bacom--adobecom.hlx.live',
   '@bacomblog': 'https://main--business-website--adobe.hlx.live',
   '@stock': 'https://main--stock--adobecom.hlx.live',
-  '@acrobat': 'https://main--acrobat--adobecom.hlx.live',
+  '@acrobat': 'https://main--dc--adobecom.hlx.live',
+  '@dc_preview': 'https://main--dc--adobecom.hlx.page',
```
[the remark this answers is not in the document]
JFernandezAdobe: Thanks! I made an update to the spec file so that our tests will run on https://www.stage.adobe.com by default.
Prod is not ready yet, so I'm also running on hlx.live for now.

features/dc_converter.spec.js
JFernandezAdobe: Done, I updated the tags. Thanks!

features/dc_converter.spec.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  name: 'DC Converter',
+  features: [
+    {
+      name: '@converter',
+      path: [
+        '/acrobat/online/pdf-to-ppt',
+      ],
+      envs: '@dc_live @dc_stage',
```
amauch-adobe, line 9: sorry, I think I may have confused you. I apologize.  We don't want to run any tests against AEM pages unless there is an absolute need for the test.  So we probably don't want to have the tag or environment '@dc_stage' unless there is a need to.  Is the dc_stage environment pointing to the milo/franklin pages for DC?  I guess I'm just trying to understand this environment better.

envs/envs.js
```diff
@@ -4,7 +4,10 @@ module.exports = {
   '@bacom': 'https://main--bacom--adobecom.hlx.live',
   '@bacomblog': 'https://main--business-website--adobe.hlx.live',
   '@stock': 'https://main--stock--adobecom.hlx.live',
-  '@acrobat': 'https://main--acrobat--adobecom.hlx.live',
+  '@dc_preview': 'https://main--dc--adobecom.hlx.page',
+  '@dc_live': 'https://main--dc--adobecom.hlx.live',
+  '@dc_stage': 'https://www.stage.adobe.com',
```
amauch-adobe, line 9: sorry, I think I may have confused you. I apologize.  We don't want to run any tests against AEM pages unless there is an absolute need for the test.  So we probably don't want to have the tag or environment '@dc_stage' unless there is a need to.  It's okay to add the prod URL since that will be necessary where it will point to your Milo/Franklin pages based on the path.   hlx.live URLs are basically the staging environment for Milo-based sites since they are generally not accessible to the public but have a level of caching just like production URLs.  That is why I don't think you will need the '@dc_stage' tag.

amauch-adobe (changes requested):
Just have a small question or concern if you could look at it when you get a chance?  Otherwise, I think it looks great.

amauch-adobe:
@JFernandezAdobe one last thing, it looks like all of your test executions failed in Nala.  I looked at the test results and it failed on the expect() call for the modal to be visible.  

Looking at the '@dc_stage' URL tag it looks like the page goes to a 404 page as shown in the screenshot below.  It looks like the environment isn't set up to house your frictionless milo-based page.  I would just remove the executions for the dc_stage tag:

Screenshot 2023-01-24 at 10 57 42 AM

The other thing I see is that on Safari the dc-converter widget block doesn't render due to some errors.  I think it may have to do with a permissions issue specific to Safari, you may want to show Blaine this if he doesn't already know:

Screenshot 2023-01-24 at 11 01 51 AM

Milo blocks will show up as a failed block on .hlx.live and .hlx.page URLs if something is wrong with the block.

Oh, and if the test keeps failing in the CI and not on your workstation you can execute your branch and test tags using the github action "Run Nala Debugging".  This will include a zip downloadable at the end of the execution that you can then unzip and use the trace files with Playwright's Trace Viewer either on your machine or here:  https://trace.playwright.dev/

I use it a lot when I'm diagnosing an issue with an execution that works fine on my machine but not in the CI pipeline.  I can help you look at this if needed.

selectors/dc_converter.selectors.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  '@pdf-converter': '.converter, .converterInline',
+  '@file-upload-input': '[data-test-id="file-upload-input"]',
+  '@pdf-complete': '//div[contains(@class,"lifecycleComplete")]',
```
Dli3, line 4: Could you not do '.lifecucleComplete' here instead of the xpath? This would locate the element faster than xpath would. 

Same for '@file-preview'.

Dli3 (approved):

envs/envs.js
JFernandezAdobe: We requested Ops to send www.stage.adobe.com requests to hlx.page and www.adobe.com to hlx.live.

Jira ticket: [WPS-21161](https://jira.corp.adobe.com/browse/WPS-21161)

JFernandezAdobe:
> Looking at the '@dc_stage' URL tag it looks like the page goes to a 404 page as shown in the screenshot below. 

Actually, the test appends a '.html' if the env is either stage or prod. The console only logs the path I added in spec.js.
Is there any way to automatically append '.html' to the test. Right now, I'm handling it like this:
` await page.goto(props.title.match(/stage|prod/) ? `${props.url}.html` : props.url);`

JFernandezAdobe:
> The other thing I see is that on Safari the dc-converter widget block doesn't render due to some errors. I think it may have to do with a permissions issue specific to Safari, you may want to show Blaine this if he doesn't already know:

> Milo blocks will show up as a failed block on .hlx.live and .hlx.page URLs if something is wrong with the block.

The issue happens in Chrome as well.
Jira ticket: [MWPW-124584](https://jira.corp.adobe.com/browse/MWPW-124584)

I will follow-up with my team regarding this issue.

amauch-adobe:
> > Looking at the '@dc_stage' URL tag it looks like the page goes to a 404 page as shown in the screenshot below.
> 
> Actually, the test appends a '.html' if the env is either stage or prod. The console only logs the path I added in spec.js. Is there any way to automatically append '.html' to the test. Right now, I'm handling it like this: `await page.goto(props.title.match(/stage|prod/) ?`${props.url}.html` : props.url);`

@JFernandezAdobe you are correct about the current code choice above.  Franklin/Milo sites generally don't use .HTML extensions, adding .HTML extensions was a hacked support from Franklin for SEO reasons.  Nala will take the paths you give it straight from the feature spec file for the paths.  If you want the URL to have .HTML ending then add it to the paths in your spec file.  Otherwise, if you still need to have the original paths that don't have the .HTML extension, I would just add it to the test code as you have already.

features/dc_converter.spec.js
JFernandezAdobe: Yes, dc_stage is pointing to the Milo pages for DC.

We requested Ops to send [www.stage.adobe.com](http://www.stage.adobe.com/) requests for the PDF to PPT page to **hlx.page**. In February, [www.adobe.com](http://www.adobe.com/) will point to hlx.live.

Jira ticket: [WPS-21161](https://jira.corp.adobe.com/browse/WPS-21161)

amauch-adobe: I'm going to resolve this conversation.  @JFernandezAdobe said he would reach out to Ops to see if stage.adobe.com is caching its pages.

amauch-adobe (approved):

amauch-adobe (changes requested):
Everything looks good, I'm happy to approve once the tests pass in the GitHub actions.  See my screenshot below all the tests are failing at this spot.  It may be a timeout/race condition on the expect() call.

Other than that, it would be good to know if stage.adobe.com is serving cached values.  Otherwise, we will still run into the caching issue I mentioned to you in slack.

Screenshot 2023-01-25 at 1 06 11 PM

JFernandezAdobe:
> Everything looks good, I'm happy to approve once the tests pass in the GitHub actions. See my screenshot below all the tests are failing at this spot. It may be a timeout/race condition on the expect() call.

Hi @amauch-adobe, as you mentioned in Slack we cannot test the pages www.stage.adobe.com in Github Action's VMs without VPN, which is not available right now. I removed @dc_stage environment from our tests.

The only remaining failure now is because of this bug: [MWPW-125063](https://jira.corp.adobe.com/browse/MWPW-125063)

image

cc @Blainegunn

JFernandezAdobe:
> The only remaining failure now is because of this bug: [MWPW-125063](https://jira.corp.adobe.com/browse/MWPW-125063)

Hi @amauch-adobe, according to Blaine, the CORS issue in MWPW-125063 should not be a problem in prod. I changed the env to @dc_prod since it seems that hlx.live will always have these kinds of issues.

amauch-adobe (approved):
Approved.

PR #56, opened 2023-01-19
Custom reporters

JingleH:
we can start using these custom reporters and generating data

i've tested the changes in my own fork of Nala [https://github.com/JingleH/nala/actions/runs/3961304265](https://github.com/JingleH/nala/actions/runs/3961304265)

API endpoints for the api-reporter are implemented in [https://git.corp.adobe.com/jinglhua/nala-api-testruns](https://git.corp.adobe.com/jinglhua/nala-api-testruns). You can also play around with it [here](https://14257-testrunsmanagement-stage.adobeio-static.net/index.html#/actions). Documentations are still in progress, and I'm thinking if and when we need to "publish" it, e.g. setting up Service Registry, CMR workflows, monitoring and CSOs, a better domain name, and jumping through hoops to get security review

Dashboard frontend that uses these APIs are also still in progress


utils/verify-specs.js
```diff
@@ -0,0 +1,39 @@
+const fs = require('fs/promises');
```
amauch-adobe, line 1: See my comments on the package.json file.  I really don't know if we need this file since a test shouldn't be added, and neither can run without having a .spec file in the features directory.  No tests should even be added/merged into the main branch without it. Plus in the future, every team will have its own repositories which would then have their .spec files located there and not in Nala's repo.  I guess I'm just trying to understand the need for this validation checker.

amauch-adobe: Is this for the UI, to ensure the titles for the reporting are correct and have what is needed?  If so, how are these going to be connected, and what is the process for the usage of this validator?

reporters/api-reporter.js
```diff
@@ -0,0 +1,48 @@
+const axios = require('axios');
+const BaseReporter = require('./base-reporter.js');
+
+// FIXME: use GitHub Actions Environment variables
```
amauch-adobe, line 4: Probably want to remove this comment unless there is still work that needs to be done in a subsequent PR to address it.

amauch-adobe: I also think maybe it would be best for the reporter's directory/files should be moved to utils rather the be top-level.  Just an opinion, not an actual requirement or anything.

amauch-adobe (changes requested):
See my comments in line.

utils/verify-specs.js
JingleH: my original intent was to catch errors fail early and not during test runs. But this is definitely not necessary, so I'll remove it

.github/workflows/manual.yml
```diff
@@ -0,0 +1,26 @@
+name: Run Nala Manually
+
+on:
+  workflow_dispatch:
+
```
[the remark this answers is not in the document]
JingleH: done!

reporters/api-reporter.js
JingleH: Originally I thought it's generally a better practice to keep configuration separate from code. Now on a second thought, since Milo will be pulling and pulling the code directly (at least it is for now), keeping non-credential and non-frequently-changing configurations in code allows consumers like Milo to not have to update its GitHub Actions Env Vars and can just run it. So let's keep it in code!

package.json
```diff
@@ -11,7 +11,8 @@
     "test:safari": "playwright test --project=webkit --headed",
     "test:chrome-headless": "playwright test --project=chromium",
     "test:firefox-headless": "playwright test --project=firefox",
-    "test:safari-headless": "playwright test --project=webkit"
+    "test:safari-headless": "playwright test --project=webkit",
+    "verify:specs": "node ./utils/verify-specs.js"
```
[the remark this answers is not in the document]
JingleH: my original intent was to catch errors fail early and not during test runs. But this is definitely not necessary, so I'll remove it

amauch-adobe (approved):
This is awesome, great work.  I approve of the changes.

PR #57, opened 2023-02-03
Playwright version update and Dependencies update

amauch-adobe:
This PR includes:

- Playwright version update 
- Dependencies updates
- Removed first keyboard focus on close button test case since per WCAG and W3 standards it is whatever the first item in the modal per order and not the close button.

No ticket was associated with this PR.


amauch-adobe:
@Dli3, when you get a chance I'm still looking into the merge to figure out why WebKit isn't playing nice with the Playwright version update.  Oddly enough, the WebKit version included with this version of Playwright didn't change and there was only a bug fix.  Version 1.29.2 on the other hand had a few more bug fixes which could have caused the issue with WebKit for keyboard input pressing with in modals.

amauch-adobe:
@Dli3 and @skumar09, 

I have logged an issue with Playwright's project here:  https://github.com/microsoft/playwright/issues/20629

There is a bug where WebKit browsers do not provide keyboard focus to the browser when the modal gets opened.  

As for now, I'm going to keep the PR open to see if we can get a resolution quickly from Playwright, but if not I will close this PR after a week.

Dli3 (approved):

PR #59, opened 2023-02-06
build automation for existing quiz page in nala

JackySun9:
https://jira.corp.adobe.com/browse/MWPW-125387

1. build automation test for existing quiz page, https://www.adobe.com/creativecloud/quiz-recommender.html?mboxDisable=1
2. try data provider with static data
3. use js-yaml to load test data
4. use npx playwright test -g @quiz-static  --config=./configs/uar.config.js --project=adobe-prod-chrome --headed to run all quiz test
5. use npx playwright test -g @result-single-template  --config=./configs/uar.config.js --project=adobe-prod-chrome --headed to run all result test


selectors/quiz.selectors.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  '@answer': '//div[@class="quiz-options"]//*[@aria-label="answer"]',
```
amauch-adobe, line 2: XPaths are available as locators for Playwright albeit it isn't the best option for finding an element.  See the documentation [here](https://playwright.dev/docs/best-practices#prefer-user-facing-attributes-to-xpath-or-css-selectors) for why it isn't the best practice.  If you can find an element using a better locator option, it's best you do.

package.json
```diff
@@ -33,6 +33,7 @@
     "@babel/eslint-parser": "7.19.1",
     "eslint": "8.31.0",
     "eslint-config-airbnb-base": "15.0.0",
-    "eslint-plugin-import": "2.26.0"
+    "eslint-plugin-import": "2.26.0",
+    "js-yaml": "^4.1.0"
```
amauch-adobe, line 37: Not a problem, just keep in mind in the future specific dependencies that will be needed for a particular test will be held under your own site's repo rather than in Nala (unless it becomes a core dependency based on Infra meetings).

features/quiz.spec.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  name: 'Quiz Recommender Blocks',
+  features: [
+    {
+      name: '@quiz',
+      path: [
+        '/creativecloud/quiz-recommender.html?mboxDisable=1',
```
amauch-adobe, line 7: This concept of adding a parameter to the path variable works fine if you just plan on having this test only ever execute against the 'main' branch. If you decide to run this test in GitHub actions against a specific PR that has a specific feature branch this will **break**.  Nala automatically will add during the parse method "?milolibs=<featureBranchName>" to the end of your URL to test the feature branch against the specific environment desired, in this case www.adobe.com.  I would probably instead add your parameter after the fact that the parse(spec.js file) has been done.  For example, ` await clickEachAnswer(`${props.url}?mboxDisable=1`, page, key);`.  You will need to add a condition to check if the URL already contains "?milolibs=" in it to decide whether to append it with an "&" or not but that will relieve the problem of it braking if used against a feature branch PR.

tests/quiz/quiz.common.js
```diff
@@ -0,0 +1,64 @@
...
+ * @param {string} url
+ * @param {import('@playwright/test').Page} page
+ * @param {string} data
+ */
+async function clickEachAnswer(url, page, data) {
+  await page.goto(url);
+
+  const answers = data.split('>').map((x) => x.trim());
+
+  // console.log(answers);
```
amauch-adobe, line 17: nitpick, probably should remove the commented-out code line.

tests/quiz/quiz.common.js
```diff
@@ -0,0 +1,64 @@
...
+  let expectProduct = results[1];
+  let acturalProduct = await page.locator(selectors['@uar-result']).textContent();
+
+  if (expectProduct === '2') {
+    // eslint-disable-next-line prefer-destructuring
+    expectProduct = results[2];
+    acturalProduct = await page.locator(selectors['@uar-result2']).textContent();
+  }
+
+  // console.log(acturalProduct);
```
amauch-adobe, line 58: nitpick, probably should remove the commented-out code line.

tests/quiz/quiz.common.js
```diff
@@ -0,0 +1,64 @@
+/* eslint-disable no-restricted-syntax */
```
amauch-adobe, line 1: word of caution, which I'm sure you have a reason for it, the point of ESLint is to help standardize code and keep it in compliance.  If we always disable it, especially for the entire file, what is the point of having it?  Milo/Nala has a strict push for compliance, it is always best to find a way to make code calls, logic, etc. that are compliant.

tests/quiz/quiz.common.js
```diff
@@ -0,0 +1,64 @@
...
+/* eslint-disable guard-for-in */
+const { expect } = require('@playwright/test');
+const selectors = require('../../selectors/quiz.selectors.js');
+
+/**
+ * @param {string} url
+ * @param {import('@playwright/test').Page} page
+ * @param {string} data
+ */
+async function clickEachAnswer(url, page, data) {
```
amauch-adobe, line 12: Question, are you planning on using the clickEachAnswer() and checkResultsPage() methods outside of the quiz test?  If so, and you have an actual case coming up, then let's create the common directory, but if not and you are just considering it in the future that may or may not happen, let's **not.** Milo/Nala follows the structure of not adding things in areas where it isn't necessary until it becomes necessary.  So, if you don't have a use case for needing these methods in a common directory, meaning there is another test that will use the exact same methods outside of the quiz method, it should be just housed in the quiz.test.js file.

tests/quiz/quiz.common.js
```diff
@@ -0,0 +1,64 @@
...
+        await page.locator(selectors['@answer'].replace('answer', options[j])).click();
+      }
+    } else {
+      // select one answer
+      await page.locator(selectors['@answer'].replace('answer', answer)).click();
+    }
+
+    if (i < answers.length - 1) {
+      // click next button
+      await page.getByRole('button', { name: selectors['@next-button'] }).click();
```
amauch-adobe, line 34: Nice usage of strict locators!

tests/quiz/quiz.test.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable no-restricted-syntax */
```
amauch-adobe, line 1: Same comment as the quiz.common.js file, let's try to find ways to be compliant rather than create exceptions.

playwright.config.js
```diff
@@ -33,7 +33,7 @@ const config = {
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: process.env.CI
     ? [['github'], ['./utils/reporters/json-reporter.js'], ['./utils/reporters/api-reporter.js']]
-    : [['list']],
+    : [['list'], ['./utils/reporters/json-reporter.js']],
```
amauch-adobe, line 36: What is the exact need for adding the json-reporter to the normal playwright executions, meaning outside of the CI/CD pipeline in GitHub?  The json-reporter is strictly necessary for the Nala Dashboard application.  I might have missed it in your code, but I didn't see it anywhere you needed this report.  Do you have a need for it?

amauch-adobe (changes requested):
Jacky Sun, this is awesome!  Way to go migrating over a test to Nala.  I left various comments and requested changes.  I think it would be best to move out your data files to a top-level data directory with a subdirectory named after your Milo-based site.  For example, nala/data/dc/. The Frictionless team handles the document cloud milo-based site, so there site is named dc.  I would do that instead of having so many directories under the tests/ directory.  Also, if you really need a common file as well, I would do the same by adding a top-level nala/common/<site-name>/ directory.

package.json
JackySun9: ok

features/quiz.spec.js
JackySun9: ok, got it

playwright.config.js
JackySun9: ok, remove it

selectors/quiz.selectors.js
JackySun9: I tried other ways, this is only way I found to make it work for all text

tests/quiz/quiz.common.js
JackySun9: Fixed it

tests/quiz/quiz.common.js
JackySun9: Fixed it

tests/quiz/quiz.common.js
JackySun9: I tried other ways, not working

tests/quiz/quiz.common.js
JackySun9: Fixed it

tests/quiz/quiz.common.js
amauch-adobe: oh, you are fine.  I actually was complementing how you chose to use the more strict locator over just a CSS selector.  So good job!

amauch-adobe (approved):
Thanks for addressing the changes @JackySun9, this looks great!  The code is approved, but before we can merge it into Nala we need to see that the tests pass in GitHub Actions.  Something you can do is use the 'Run Nala Debugging' GitHub Action Workflow, you will find it under the actions tab in GitHub under your forked repository URL page.  This workflow just needs you to tell it what branch to run Nala on, the default is 'main', and then supply it a test tag from your quiz.spec.js file, my guess is you will want to use '@quiz'.  Let it run and then you can download the playwright report artifact and upload the trace files for each test iteration at the Playwrights Trace Viewer [here](https://trace.playwright.dev/)

skumar09 (commented):
@JackySun9 , 
- would like to understand bit more on the test data files (.yml), and see if there is any other way to feed the data? 
- can we sync up sometime when you are free?..

Dli3 (approved):

amauch-adobe:
After having some slack conversations with Jacky Sun, it turns out this is a POC for right now while his team is creating the actual Milo/Franklin-based pages.  Currently, the test executes against AEM pages, we do not want to use Nala to test AEM pages.  The Nala Automation framework is meant to test Milo blocks and Milo-based consumer sites.  For this reason, I felt it is best to hold off merging this into Nala until the test/PR is adjusted to use real Milo/Franklin pages and not AEM pages especially since it might change test timeouts, parallel executions, etc. due to the performance of Milo vs. AEM.

cc @Dli3, @skumar09

features/uar/quiz.spec.js
```diff
@@ -0,0 +1,77 @@
+module.exports = {
+  name: 'Quiz Recommender Blocks',
+  features: [
+    {
+      tcid: '1',
```
amauch-adobe, line 5: I think tcid's start at 0?  @skumar09 can clarify if it is required or not though.

features/uar/result.spec.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  name: 'Quiz Recommender Blocks',
+  features: [
+    {
+      tcid: '1',
```
amauch-adobe, line 5: I think tcid's start at 0?  @skumar09 can clarify if it is required or not though.

libs/data-provider.js
```diff
@@ -0,0 +1,23 @@
+const fs = require('fs');
+// eslint-disable-next-line import/no-extraneous-dependencies
+const yaml = require('js-yaml');
+const { request } = require('@playwright/test');
+
+/**
```
amauch-adobe, line 6: It would be good to add a method description so when used in an IDE it will pick it up for auto-complete.

libs/data-provider.js
```diff
@@ -0,0 +1,23 @@
...
+ * @param {string} filePath
+ */
+async function loadTestData(dataFilePath) {
+  return dataFilePath.includes('.yml') ? yaml.load(fs.readFileSync(dataFilePath, 'utf8')) : JSON.parse(fs.readFileSync(dataFilePath, 'utf8'));
+}
+
+/**
+ * @param {string} path
```
amauch-adobe, line 14: It would be good to add a method description so when used in an IDE it will pick it up for auto-complete.

selectors/uar/quiz.page.js
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * click get your results button
+   */
+  async clickResultButton() {
+    await this.resultButton.click();
+  }
+
+  /**
+   * @param {string} url
```
amauch-adobe, line 39: Missing method description

selectors/uar/quiz.page.js
amauch-adobe, line 1: Just a reminder on CSS selectors and XPath locators, performance wise they take the longest.  If you can use ids or class names, etc. that is preferred.

selectors/uar/quiz.page.js
```diff
@@ -0,0 +1,96 @@
...
+        await this.clickNextButton();
+      }
+    }
+
+    // click get your results button
+    await this.clickResultButton();
+  }
+
+  /**
+   * @param {string} result
```
amauch-adobe, line 71: Missing method description.

selectors/uar/result.page.js
amauch-adobe, line 1: Same case, just a reminder on CSS/XPath selectors in Playwright they are not as performant as they provided locators.  https://playwright.dev/docs/locators

amauch-adobe, line 1: POM methods should have method descriptions for auto-complete usage in IDEs.

selectors/uar/result.page.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /**
+   * @param {string} url
+   * @param {import('@playwright/test').Page} page
+   * @param {string} results
+   */
+  async checkResultPage(url, results) {
+    await this.page.goto(url);
+    await this.page.reload();
```
amauch-adobe, line 28: Do you need to refresh the page right after going to the page?  Just curious, not a concern, just wondering.  Maybe add a note if it is needed.

selectors/uar/result.page.js
```diff
@@ -0,0 +1,98 @@
...
+
+      if (locator.includes('social-ad')) {
+        await this.socialAd.click();
+      }
+
+      if (locator.includes('flyer')) {
+        await this.flyer.click();
+      }
+
+      await this.page.waitForTimeout(1000);
```
amauch-adobe, line 88: Just a recommendation (not a requirement), when I add basically a timeout like this (similar to a thread sleep or an implicit/explicit wait) it helps to explain why since Playwright recommends not to use them unless absolutely needed.

selectors/uar/result.page.js
```diff
@@ -0,0 +1,98 @@
...
+      expect(this.page.url()).toContain(`${value.split(':')[2]}`);
+      await this.page.goto(url);
+      await this.page.reload();
```
amauch-adobe, line 91: Same case as the previous reload call it would be best to explain the need for calling it.

tests/uar/result.test.js
```diff
@@ -0,0 +1,32 @@
...
+import { Result } from '../../selectors/uar/result.page.js';
+
+const { test } = require('@playwright/test');
+const ResultSpec = require('../../features/uar/result.spec.js');
+
+const { features } = ResultSpec;
+const { loadTestData } = require('../../libs/data-provider.js');
+
+test.describe('Result flow test suite', () => {
+  // test.setTimeout(10 * 60 * 1000);
```
amauch-adobe, line 12: small nitpick if the test timeout is no longer necessary, we could remove the code and comment.

tests/uar/result.test.js
```diff
@@ -0,0 +1,32 @@
...
+        const result = new Result(page);
+        const url = `${baseURL}${feature.path}`;
+
+        // load test data from static files
+        const testdata = await loadTestData(`${feature.data}`);
+
+        for (const key of Object.keys(testdata)) {
+          const resultUrl = url + key;
+          console.info(resultUrl);
+          page.waitForTimeout(5000);
```
amauch-adobe, line 26: Good to explain why we need a 5sec timeout here.

tests/uar/quiz.test.js
```diff
@@ -0,0 +1,38 @@
...
+import { Quiz } from '../../selectors/uar/quiz.page.js';
+
+const { test } = require('@playwright/test');
+const QuizSpec = require('../../features/uar/quiz.spec.js');
+
+const { features } = QuizSpec;
+const { loadTestData } = require('../../libs/data-provider.js');
+
+test.describe('Quiz flow test suite', () => {
+  test.setTimeout(3 * 60 * 1000);
```
amauch-adobe, line 12: Good to explain the need for a 3min test timeout for clarification, especially for future purposes of test performance audit in the CI execution.  Just an idea.

amauch-adobe (changes requested):
Just a few comments, but well done, I like the test a lot!

amauch-adobe:
@JackySun9 it looks like your test uses the baseUrl construct which requires having a config.js in the /configs directory.  Are you using the acom.config.js file to pull your baseUrl?  If so we will need to use the specific label on your PR so the Nala CI knows which config to use for your test execution.

JackySun9:
> 

yes, thanks!

selectors/uar/result.page.js
JackySun9: Got it, thanks!

JackySun9:
@amauch-adobe thanks for your feedback, I resolved all your concerns

features/uar/quiz.spec.js
Dli3: Might be good to clarify this because he started at 1 in [the carousel test](https://github.com/adobecom/nala/blob/main/features/milo/carousel.block.spec.js). 
@amauch-adobe @skumar09 @JackySun9

skumar09:
@JackySun9  Welcome to Nala, !! Nice to see UAR PR 
provided few review inputs, pls review and update.

skumar09 (changes requested):

JackySun9:
@skumar09 I updated them according to your feedback, thanks!

amauch-adobe (approved):
Looks great!

Dli3 (approved):

PR #69, opened 2023-02-21
[MWPW-126307] Added first FEDS consent check

chadsunvice:
Current status: Eurovision is working on adding new FEDS features into our feature-branch. 
Current checks being added run on feds-gnav branch.


.gitignore
```diff
@@ -117,3 +120,4 @@ screenshots
 /test-results/
 /playwright-report/
 /playwright/.cache/
+Test.pptx
```
amauch-adobe, line 123: I'm going to reach out to Frictionless to have an actual directory created to test downloads so we can ignore the directory rather than keep adding each file to the ignore list.  For now, though this will work.

envs/envs.js
```diff
@@ -1,9 +1,11 @@
 module.exports = {
-  '@milo_live': 'https://main--milo--adobecom.hlx.live',
-  '@milo_prod': 'https://milo.adobe.com',
-  '@bacom_live': 'https://main--bacom--adobecom.hlx.live',
-  '@bacomblog_live': 'https://main--business-website--adobe.hlx.live',
-  '@stock_live': 'https://main--stock--adobecom.hlx.live',
+  '@milo': 'https://main--milo--adobecom.hlx.live',
```
amauch-adobe, line 2: Please don't change the naming constructs of the envs.js for the tag names.  All names either have to have _preview for hlx.page, _live for hlx.live, or _stage for stage.<domain_name>, and _prod for production domain urls.  Our parse.js requires that construct to correctly create urls for testing on PRs as well as the Nala Dashboard Test Results application coming soon.  You might just needed to update your fork to the latest main.

amauch-adobe: Also, as a word of caution the hlx.page URLs are not behind any caching so they will be more taxing against the Franklin services.  For test development, authoring, and engineering development hlx.page URLs are perfect for those types of actions.  As soon as the page under test is published it is best to use hlx.live URLs instead or the actual production URLs.  hlx.live URLs are behind caching but may not have all of the other services running against them, for example, chat.  If you are testing a page that is located under a /drafts/ directory in your site's folder in Sharepoint, the risk level of harm for having that page published under drafts is so small since we filter out those pages being indexed or let alone connected to any level of path caching in Akamai or CDNs that the likelihood of someone going to that page other than internal employees is very small.  That is why Milo's block testing is done under the drafts directory and has pages already published so that we can use hlx.live URLs and tax the Franklin services less due to caching.

features/consent.spec.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  name: 'FEDS Consent',
+  features: [
+    {
+      name: '@consent',
+      path: [
+        '/libs/feds/drafts/qa/consent/feds-consent-page',
+      ],
+      envs: '@feds',
```
amauch-adobe, line 9: will need to change envs to '@feds_live'

amauch-adobe (changes requested):
@chadsunvice this is an excellent test, good job on your first test using Nala/Playwright.  I just had a few comments and one specific requested change that needs to be made before this can be merged.  Once again, great job!

tests/consent.test.js
```diff
@@ -0,0 +1,41 @@
...
+      const OneTrustModalDontEnable = page.locator(selectors['@OneTrustDontEnable']);
+      const OneTrustModalEnableAll = page.locator(selectors['@OneTrustEnableAll']);
+      const OneTrustModalClose = page.locator(selectors['@OneTrustModalClose']);
+
+      // Load OneTrust consent component page:
+      const url = props.title.match(/stage|prod/) ? `${props.url}.html` : props.url;
+      await page.goto(url);
+
+      // Wait for the OneTrust consent bar to be displayed:
+      await page.waitForSelector(selectors['@OneTrustContainer']);
```
JackySun9, line 30: For consent banner, it will only show for some locales, do you need mock GEO for it?

JackySun9 (changes requested):

tests/consent.test.js
narcis-radu: Good point @JackySun9 ! We developed some helpers to allow mocking geo location, @chadsunvice can introduce them and make sure the consent banner is displayed.

selectors/consent.selectors.js
```diff
@@ -0,0 +1,17 @@
...
+  '@OneTrustCookiesButton': '#onetrust-pc-btn-handler',
+  '@OneTrustDontEnableButton': '#onetrust-reject-all-handler',
+
+  // OneTrust container selectors:
+  '@OneTrustContainer': '#onetrust-banner-sdk',
+  '@OneTrustConsentFrame': '#onetrust-pc-sdk',
+  '@OneTrustConfirmChoices': 'button.save-and-close',
+  '@OneTrustModalDontEnable': 'button.disable-all-btn',
+  '@OneTrustModalEnableAll': 'button.enable-all-btn',
+  '@OneTrustModalClose': 'button.save-and-close'
```
narcis-radu, line 16: Looks like this selector `button.save-and-close` is duplicated. We should update the value to `#close-pc-btn-handler`. Generally speaking, it would be a good idea to have specific selector and make sure we don't select unwanted elements (e.g. button.save-and-close might be the button from a random form).

tests/consent.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+      const OneTrustContainer = page.locator(selectors['@OneTrustContainer']);
+      const OneTrustConsentFrame = page.locator(selectors['@OneTrustConsentFrame']);
+      const OneTrustConfirmChoices = page.locator(selectors['@OneTrustConfirmChoices']);
+      const OneTrustModalDontEnable = page.locator(selectors['@OneTrustDontEnable']);
+      const OneTrustModalEnableAll = page.locator(selectors['@OneTrustEnableAll']);
+      const OneTrustModalClose = page.locator(selectors['@OneTrustModalClose']);
+
+      // Load OneTrust consent component page:
+      const url = props.title.match(/stage|prod/) ? `${props.url}.html` : props.url;
```
narcis-radu, line 26: I don't think we have use-cases for pages with `.html` extension 🤔

tests/consent.test.js
narcis-radu: `?customPrivacyLocation=de` should force the banner to show up.

tests/consent.test.js
```diff
@@ -0,0 +1,41 @@
...
+      await page.goto(url);
+
+      // Wait for the OneTrust consent bar to be displayed:
+      await page.waitForSelector(selectors['@OneTrustContainer']);
+      await expect(OneTrustContainer).toBeVisible();
+
+      // Check the contents of the consent bar:
+      await expect(OneTrustTitle).toHaveText('Make It Your Own');
```
narcis-radu, line 34: Title will be different based on locale, maybe we shout expect for a DOM element with class <x> that holds the title text; the same for other expectations.

.gitignore
chadsunvice: I agree @amauch-adobe. I'll keep as-is and we can remove those in the future.

envs/envs.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
   '@milo_live': 'https://main--milo--adobecom.hlx.live',
-  '@milo_prod': 'https://milo.adobe.com',
+  '@milo_prod': 'https://milo.adobe.com/',
```
narcis-radu, line 3: The extra `/` added might break things

tests/consent.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+      const OneTrustContainer = page.locator(selectors['@OneTrustContainer']);
+      const OneTrustConsentFrame = page.locator(selectors['@OneTrustConsentFrame']);
+      const OneTrustConfirmChoices = page.locator(selectors['@OneTrustConfirmChoices']);
+      const OneTrustModalDontEnable = page.locator(selectors['@OneTrustDontEnable']);
+      const OneTrustModalEnableAll = page.locator(selectors['@OneTrustEnableAll']);
+      const OneTrustModalClose = page.locator(selectors['@OneTrustModalClose']);
+
+      // Load OneTrust consent component page:
+      // !Note: OneTrust only loads on a specific subset of locales.
```
narcis-radu, line 27: The banner gets loaded only on GDPR enforced countries.

envs/envs.js
narcis-radu: @chadsunvice - not sure I understand the 👎 so I will try to add more context: in your PR you added an extra `/` to `@milo_prod` and I'm concerned it will break things, the end URL will be something like `milo.adobe.com//path/to/page`. I checked the file in `main` branch and it does not have the extra `/`. I don't think this is an update you inherited from merging.
Here's the file in the main branch - https://github.com/adobecom/nala/blob/main/envs/envs.js#L3

features/parse.js
```diff
@@ -19,8 +19,9 @@ function buildUrl(url, env) {
       branch = branchBuild.slice(0, (branchBuild.lastIndexOf('-')));
     }
   }
+
   if (!repoName) {
-    repoName = 'milo';
+    repoName = 'milo_live';
```
JackySun9, line 24: milo repo name is milo, right?

envs/envs.js
amauch-adobe: yeah, none of the environment domains in envs.js should have a trailing forward slash at the end since the path variable in the feature.spec file has the forward slash added to the URLs there.

features/parse.js
amauch-adobe: yeah, there shouldn't be any change to the parse.js file being made in test PRs.  Where that is a core Nala script file any changes should be made separately outside of additional test PRs.  And @JackySun9 you are correct it should just be 'milo' and NOT 'milo_live'.

amauch-adobe: So this shouldn't be changed.  Lets revert the commit for this line so it stays as repoName = 'milo'.  The actual repository name comes from github's API during the github actions workflow, it sends the entire short name of the repository that executes the github action, i.e., adobecom/milo or if for 'DC' it would grab adobecom/dc.  We use this naming to decide how the feature branch that is being tested should be added to the url, either ?milolibs=featurebranch or replacing the "main" in the hlx.live url to the feature branch name.

amauch-adobe (changes requested):
Just a requested revert to the features/parse.js.

narcis-radu (approved):

PR #71, opened 2023-02-21
Removed workaround for CSP issue

amauch-adobe:
Removed workaround for CSP issue with DC hlx.* urls when testing for Milo or any other consuming site but DC's repository.


JFernandezAdobe (approved):
Approved

PR #72, opened 2023-02-21
Review block smoke test

JFernandezAdobe:
Jira ticket: [MWPW-126770](https://jira.corp.adobe.com/browse/MWPW-126770)

Smoke test for the Review block:

- Component title should be visible
- The ratings field is visible
- The stats info from Review API is present


JFernandezAdobe:
Updated the spec file to use only one tag, `@frictionless`. Using two different tags `@review` and `@converter` runs the same test twice.

image


cc @JackySun9 @amauch-adobe

JackySun9 (approved):

selectors/review.selectors.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  '@review-block': '[class="review"]',
```
amauch-adobe, line 2: does this not work with just having the value of `'@review-block':'.review'`?  If it does, this is more of a nit pick, why not just do that than do an XPATH.  The same is for all the other selectors finding things by class names.

tests/review.test.js
```diff
@@ -0,0 +1,30 @@
...
+
+      await expect(reviewBlock).toBeVisible();
+      if (await failedBlock.isVisible()) {
+        console.log(`${browser.browserType().name()}: ${await failedBlock.getAttribute('data-reason')} on ${props.url}`);
+        await expect.soft(failedBlock).not.toBeVisible();
+      }
+      await expect(reviewTitle).toBeVisible();
+      await expect(reviewStats).toBeVisible();
+      await expect(ratingFields).toBeVisible();
+    });
```
amauch-adobe, line 28: This obviously is just a comment/opinion, but should the test also validate that a star when clicked is saved/marked and validated end-to-end?  Maybe a test case expect for checking when a star is clicked it is marked/active, and then when it is unclicked it is not marked/active.  Just thinking out loud, it's your team's test, so no worries here.

amauch-adobe (approved):
Approved, just made some comments/questions that might be worth looking at.  As the test is, it works and is approved.  Just as a side note/reminder, using XPATH in Playwright takes longer for an element to be found, it still is available, but if you can find an element using a different locator it is best.

amauch-adobe:
@JFernandezAdobe, I'm okay with having this merged, but I was curious if you saw my comments on the PR.

tests/review.test.js
JFernandezAdobe: Added additional checks. Thanks!

JFernandezAdobe:
I've added more checking and updated all our tests to run on hlx.live.
Thanks for the review @JackySun9 and @amauch-adobe!

selectors/review.selectors.js
JFernandezAdobe: Sorry, I missed this. I'll update this on my next PR.

PR #75, opened 2023-03-08
Do not run DC tests in hlx.live

JFernandezAdobe:
Resolves: [MWPW-127629](https://jira.corp.adobe.com/browse/MWPW-127629)

Due to the huge number of errors logged when testing feature branches in Milo using milolibs parameter (e.g. https://github.com/adobecom/milo/pull/497), we should not run the Frictionless tests in hlx.live anymore.

image

If we don't use the milolibs parameter it is fine or if you leave it empty, it is fine and works as normal.

All `@dc_live` tags in this PR are replaced with `@adobe_stage` since the milolibs param does not work in stage.adobe.com.


amauch-adobe (approved):
Approved.

JackySun9 (approved):

JFernandezAdobe:
@amauch-adobe I will close this PR for now since it is dependent on these two tickets to be completed first:


[MWPW-128229](https://jira.corp.adobe.com/browse/MWPW-128229) CircleCI Research and integration

[MWPW-128233](https://jira.corp.adobe.com/browse/MWPW-128233) Update Parse.js to add milolibs?= parameter to stage branch urls

PR #76, opened 2023-03-08
MWPW-127667 Update playwright configuration

skumar09:
MWPW-127667 : Update Nala playwright configuration for following requirements

- Parallel run 
    -  Parallel run is enabled but number of workers count parameter was not updated.
    -  Updated workers count = 2  
- Provide path for test-results folder
   - Added the test result path
- Provided path for local -test debug html results
   - Added path for html test results for local env debugging 
   - This doesn't have impact on github action or CI execution
- Updated actionTimeout to 30 sec. ( It was at 0 i.e. 'No limit')


JingleH:
One thing I would suggest is that if we want to change the html output location, we should also add it to .gitignore

playwright.config.js
```diff
@@ -29,15 +30,15 @@ const config = {
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  workers: process.env.CI ? 2 : undefined,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: process.env.CI
     ? [['github'], ['./utils/reporters/json-reporter.js'], ['./utils/reporters/api-reporter.js']]
-    : [['list']],
+    : [['html', { outputFolder: 'test-html-results' }]],
```
JackySun9, line 37: can we have both list and html?

skumar09:
> One thing I would suggest is that if we want to change the html output location, we should also add it to .gitignore
@JingleH , good point, i have updated the PR, pls review

JackySun9 (approved):

JingleH (approved):

playwright.config.js
JingleH: seconding this. But I'm also thinking that do people want that generated html by default? It does increases the execution time at the end by quite some lot

playwright.config.js
```diff
@@ -29,11 +30,11 @@ const config = {
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  workers: process.env.CI ? 2 : undefined,
```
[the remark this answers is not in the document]
amauch-adobe: I'm cool with trying it out and seeing how it fairs over the longterm, as I mentioned I tried it in one PR and kept updating it over the day to see test results.  Longterm might be a good test factor to see if it was just an issue for the time I was doing it.  Let's give it a try, we can always discuss changing it later or something.

.gitignore
```diff
@@ -118,6 +118,7 @@ debug
 reports
 screenshots
 /test-results/
+/test-html-results/
```
amauch-adobe, line 121: This entirely depends on if we decide to always print out the HTML report for local runs, see my comment below.  If we are going to then yes let's add this to .gitignore, otherwise all HTML reports will by default go to the playwright-report/ directory that is already in the .gitignore file.

amauch-adobe (changes requested):
Love the idea of adding parallel executions, see my comments below.

playwright.config.js
```diff
@@ -29,11 +30,11 @@ const config = {
...
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: process.env.CI
     ? [['github'], ['./utils/reporters/json-reporter.js'], ['./utils/reporters/api-reporter.js']]
-    : [['list']],
+    : [['html', { outputFolder: 'test-html-results' }]],
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
```
amauch-adobe, line 40: I saw in a previous commit that the actionTimeout was updated to 300000 milliseconds, did you revert the change?

playwright.config.js
skumar09: @amauch-adobe  this html report is only in local laptop/machine run, not in git. this will reduce lot of effort in test debug locally.

amauch-adobe (changes requested):

playwright.config.js
skumar09: - yeah , i was thinking failure for FaaS script was due to timeout, but looks their failure was different issue. 
- I will not encourage to have actionTimout set to 0 i.e. no limit.  we should set somewhere between 60 sec to 300 sec.
-

amauch-adobe:
@skumar09 and I had a quick huddle on my comments to the PR and worked through them. All are approved after we decided to add back the actionTimeout to 30 sec.

amauch-adobe (approved):
Approved.

PR #79, opened 2023-03-10
[MWPW-126969] Added search component check & page-objects

chadsunvice:
**WHY?** 
Moving forward, there will be a need to reuse code written by other NALA clients (_eg: IMS login methods_) and limit the test scope to a single testable component, while also reducing the test file foot-print (_selectors should be defined in page-objects and imported on-demand_).

**HOW?**
Each NALA client should create a base page that keeps the common client login (_eg: feds.page.js_), that is then inherited by any page-object (_eg: feds.header.page.js_), which represents a single, atomic, testable component in the AUT (_selectors, props, methods only!_).

**WHAT?**
By leveraging the power of PlayWright POM. This PR represents a scaled-down model, which tests the FEDS GNAV search component.

**WHAT NOT?**
There is a reason why multiple inheritance is not allowed in JS and other higher level programming languages! I first considered using a _**section > page >> page1 | page2 | page3 >> (multiple inheritance) specific-page**_ approach, but seeing how cluttered and unmaintainable the Platform-UI variant has become by using polytype mixins is something I think we don't think would be desirable in NALA.

The end-result are large, cluttered pages which inherit from multiple client sections/pages and become unreadable and unmaintainable. Thus, every client should create their own page-objects on-demand, following this simple pattern.


pages/feds.page.js
```diff
@@ -0,0 +1,9 @@
+import { expect, test } from '@playwright/test';
+
+exports.FedsPage = class FedsPage {
+
+  constructor() {
+    // >> Generic FEDS page selectors declared here <<
```
JackySun9, line 6: If we need login, where to put it

chadsunvice: I see login inside a `ims.page.js` file, as it is not specifically a FEDS-related feature, but more as a global service which every consumer needs to use, on-demand (_some tests require login, others don't_).

This PR tries to organize our FEDS logic a bit better and have custom methods tailored around specific actions (eg: `openUserProfile`, `closeUserProfile`, `openAppLauncher`, `closeAppLauncher`) & have the selectors in specific files, as to not pollute the test-file.

I will add the `ims.page.js` logic if way of working is approved. For now, I am going by the logic already available (Footer/Header/Search/Consent). Login would be next after I add the Header/Footer checks. The selectors added are required for the operations to test the GNAV.

JackySun9 (approved):

tests/search.test.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable no-restricted-syntax */
+/* eslint-disable no-await-in-loop */
+const parse = require('../features/parse.js');
+const search = require('../features/search.spec.js');
+const { expect, test } = require('@playwright/test');
```
amauch-adobe, line 5: It's a lame nitpick and opinion, but I like having the playwright/test import be at the top and the rest of the imports from our own framework below.  Just a personal preference, feel free to make your own decision on that.

pages/feds.page.js
amauch-adobe, line 1: This type of template file might be best housed under the /test-examples/templates directory or maybe a /test-templates directory.  @skumar09 what do you think?

amauch-adobe: wait it looks like you were starting on this type of page and just added comments like it was a template since it is just an empty constructor.  Sorry, I first thought this was just a template of sorts, if it is intended to be used as a template I would say go for having it under a different directory for learning.

pages/feds.header.page.js
```diff
@@ -0,0 +1,127 @@
...
+
+exports.FedsHeader = class FedsHeader extends FedsPage {
+
+  constructor(page) {
+    super();
+    this.page = page;
+    this.page.props = {
+      lsKeyName: 'feds-notifications_en',
+      components: {
+        'NavBar': '#feds-topnav',
```
amauch-adobe, line 12: Silly question, but just want to know why these component selectors are under the constructor instead of below it like all the other selectors.  I also didn't see anywhere the headerData {}, or pageClouds: [] data structures are used.  What was your plan there?

amauch-adobe (commented):
See my comments throughout.  That being said this looks amazing, cool, clean, and susinct.  Having POMs hold selectors though starts to beg the question of why we have the .selectors.js files at all.  Should POMs take over selectors.js files?  Should POMs be used when needed for page reuse only and selectors are then added there instead of the .selectors.js files and .selectors.js files are used only when testing blocks or a quick page that won't be reused?  

I think this is why this type of discussion should be pushed to the http://milo.adobe.com/community board for intake.  Then after things are fledged through we can either adjust the PR if needed or not and then merge and approve.  What do you think @skumar09?

tests/search.test.js
chadsunvice: It is not nitpick, I agree as well @amauch-adobe . I made the changes and pushed updated version.

pages/feds.page.js
chadsunvice: Yes, this is not a template, it is meant to house FEDS specific logic across all components in the future (_eg: Header/Footer/Search/IMS/Consent/Jarvis/etc_). Different clients should create their own inheritance schema on-demand.

pages/feds.header.page.js
chadsunvice: I agree these were confusing, so I removed them. I can explain the use-case though. These are not meant to be page-selectors, but more like FEDS component-to-UI-element mappings. I can explain their use below.

In our FEDS browser objects, we might have pages where certain components are, or aren't enabled, via client requirements (_eg: Some pages might have SubNav bar, some might have only NavBar, some might have MegaMenu. Some pages might have SkinnyFooter, some might have the regular Footer component. We don't know until we check the feds.components object keys_).

Thus, we need a collection of sorts to be able to iterate through the "component dictionary". An example below:

```
    // Looping through page-enabled FEDs components:
    Object.keys(feds.components).forEach((component) => {
      const componentElem = $(Header.props.components[component]);
      console.info(`[EuroInfo] '${component}' component display status: ${componentElem.isDisplayed()}`);
      switch (component) {
        case 'subnav':
          if (pageHasSubnav) {
            expect(componentElem.isDisplayed(), `Asserting if [${component}] is displayed`).to.be.true;
          } break;
        case 'breadcrumbs':
          if (pageHasBreadcrumbs) {
            expect(componentElem.isDisplayed(), `Asserting if [${component}] is displayed`).to.be.false;
          } break;
        case 'regionPicker':
          if (pageHasRegionPicker) {
            expect(componentElem.isDisplayed(), `Asserting if [${component}] is displayed`).to.be.true;
          } break;
        default:
          expect(componentElem.isDisplayed()).to.be.true;
          break;
      }
    });
```

I removed the props and will add them when those specific methods need to be added.

amauch-adobe (approved):
I approve of the approach and the changes.  That being said, the merge of the POM model should wait until @skumar09 and indeed a discussion in the Milo community forum can be done over the selectors.js files usage, and need against POMs and selectors.  That is my feeling.

amauch-adobe:
I added the do not merge label until @skumar09 and the Milo community can discuss the approach for selectors.js files in conjunction with POMs.

pages/feds.page.js
skumar09: - As discussed in past few Nala discussion about including page object into current structure, i think we are good on that. On the naming convention I would still stick to selector kind [<block or section or any name>.selector.js] as it will provide kind of uniformity. [open for ideas/thoughts..]
- We can move reusable functions or methods under libs 
- Please update the folder/file path as per recent changes with respect to PR#88

skumar09 (changes requested):

pages/feds.page.js
chadsunvice: Thanks for the review @skumar09! Can you please expand on the `i think we are good on that`? I am looking for more information to understand the though process. Can you please link the the NALA thread from the Milo Community folder in which this has been discussed and agreed upon?

AFAIK, the PlayWright core team suggests making use of POM w/e possible. I would want to understand the though process about not breaking up logic of clients per page, as opposed to having bloated libs. Here is a nice walkthrough of the process in case you are not familiar with how it works: https://www.lambdatest.com/learning-hub/playwright-page-object-model.

**The benefits are as follows:**

1. Modular approach to structuring our information on a per page/per client basis
2. Greatly improving readability (_Selectors and logic added in the test-file distracts from the test functionality._)
3. Greatly reducing the test-file footprint. We define key-value pairs in `<feature>.selectors.js`, then we add them to the current working page in the test-file. Why? If we need to update with a new selector, there are 2 places wwhere we need to make the change, instead of one, in the page-object. (_I don't know if this is scalable, or idiomatic tbh._)

eg:

 ```
test.describe(`${name}`, () => {
  features.forEach((props) => {
    test(props.title, async ({ page, browser }) => {
      const { url } = props;
      const pageNameRegex = /(?<=online\/)([^.?]+)/gi;
      const input = fileInputList.filter((x) => x.pages.includes(url.match(pageNameRegex)[0]))[0];
      const converterBlock = page.locator(selectors['@pdf-converter']);
      const fileInput = page.locator(selectors[input.locator]);
      const exportButton = page.locator(selectors['@export-convert-button']);
      const convertButton = page.locator(selectors['@convert-button']);
      const insertButton = page.locator(selectors['@insert-button']);
      const plusButton = page.locator(selectors['@plus-button']);
      const mergeButton = page.locator(selectors['@merge-button']);
      const inputPassword = page.locator(selectors['@input-password']);
      const confirmPassword = page.locator(selectors['@confirm-password']);
      const setPassword = page.locator(selectors['@set-password']);
      const pdfComplete = page.locator(selectors['@pdf-complete']);
      const filePreview = page.locator(selectors['@file-preview']);
      const protectHeading = page.locator(selectors['@protect-heading']);
      const lowCompress = page.locator(selectors['@low-compress-option']);
      const downloadButton = page.locator(selectors['@download']);
      const failedBlock = page.locator(selectors['@widget-block-failed']);
```

4. Giving NALA clients the freedom to structure their logic on their own reusable page-objects

*Cc:* @amauch-adobe & @auniverseaway.

skumar09: @chadsunvice 
- It is goes back to multiple initial Nala discussions (which doesn't have Milo threads) where we discussed of having page 
  object classes for selector.js files, and also try to keep Nala as lite as possible. 
- On playwright suggestion about POM and its benefits, I am aware of it, and its one of the good approach. 
- On my point  about naming convention , it is using of verbiage `page vs feature.selector.js.`  And `<feature>.selector.js `would be updated to page(s) of POM, not one more separate file
- Also we have to enhance/update current parse.js, to accommodate some of these and recent changes. ( I will be raising a jira for same )
cc : @amauch-adobe @auniverseaway

chadsunvice: Hello @skumar09, @amauch-adobe, @auniverseaway! I maybe didn't made myself understood, when I said that POM will make the NALA framework clean and efficient, as opposed to hardcoded and prone to future scalability issues. So take a look at: https://github.com/adobecom/nala/pull/90. I tried to show a side-by-side comparison of how we can re-organise the logic as to make the test-files tackle only higher level logic, and keep functionality & selectors bound to the POs.

chadsunvice:
Hello @skumar09, @amauch-adobe, @auniverseaway! I maybe didn't made myself understood, when I said that POM will make the NALA framework clean and efficient, as opposed to hardcoded and prone to future scalability issues. So take a look at: https://github.com/adobecom/nala/pull/90. I tried to show a side-by-side comparison of how we can re-organise the logic as to make the test-files tackle only higher level logic, and keep functionality & selectors bound to the POs.

amauch-adobe:
I'm good with the approach, I also added a discussion in the milo community here.  https://github.com/orgs/adobecom/discussions/589. I also think POMs should use the naming and folder structure as proposed in #90.  I'm okay moving forward.

amauch-adobe:
@chadsunvice you will want to make a small adjustment to your import for the search test's import of the parse.js since it has been moved to a different location now.  See how Nala is failing to run since the import fails below:

Screenshot 2023-03-27 at 6 05 08 PM

chadsunvice:
> @chadsunvice you will want to make a small adjustment to your import for the search test's import of the parse.js since it has been moved to the libs folder now. See how Nala is failing to run since the import fails below:
> 
> Screenshot 2023-03-27 at 6 05 08 PM

I pulled the latest changes Santosh's PR and updated with the requests made in the UserProfile PR.

amauch-adobe:
> > @chadsunvice you will want to make a small adjustment to your import for the search test's import of the parse.js since it has been moved to the libs folder now. See how Nala is failing to run since the import fails below:
> > Screenshot 2023-03-27 at 6 05 08 PM
> 
> I pulled the latest changes Santosh's PR and updated with the requests made in the UserProfile PR.

Thanks @chadsunvice everything looks good from the Nala executions, I'm good going forward and merging.

cc @skumar09

amauch-adobe (approved):
Approved.

skumar09:
look good, one small input: Using camelCase for identifier names (variables and functions) [ similar to PR #90 ]

chadsunvice:
> Approved.

Updated according to @skumar09! Updated the other PRs as well to match.

chadsunvice:
> look good, one small input: Using camelCase for identifier names (variables and functions) [ similar to PR #90 ]

Updated! The variables like `fedsConfig` will remain like that as to match the actual browser-object naming. Some PRs have such variables. But for the scope of the test-files and POs, I updated according.

amauch-adobe (approved):
Approved.

JackySun9 (approved):

PR #84, opened 2023-03-16
HTML extension test uses locale urls

amauch-adobe:
This PR addresses the found issue in this ticket: [MWPW-127111](https://jira.corp.adobe.com/browse/MWPW-127111)

* Added regex checking for HTML extensions when using locales in /blog/ URLs
* Added JP URL to test HTML extension in a different locale.

### DIFF
--- a/features/html_ext.spec.js
+++ b/features/html_ext.spec.js
@@ -9,6 +9,7 @@ module.exports = {
         '/customer-success-stories/abb-case-study',
         '/customer-success-stories/dentsu-isobar-case-study',
         '/blog/',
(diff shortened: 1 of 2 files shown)


amauch-adobe:
The Nala executions were canceled because they fired off a lot of executions instead of just one.  It also pulled in the trivial label and I think that caused the GitHub actions job to wig out.  Anyways, I re-ran it manually in GitHub Actions using the Nala Debugging workflow and it passed as shown below:

Screenshot 2023-03-16 at 1 23 22 PM

PR #90, opened 2023-03-23
[MWPW-128180] Added FEDS user-profile checks 

chadsunvice:
Federated Services wants to use this PR to show-case the power of POM (_Done right!_). We will use this PR as a working example for a future **#milo-community** post regarding how we write our checks.

For this example, we will try to automate the **FEDS user profile** component, in a dynamic, scalable, idiomatic way. Please compare the FEDS login checks approach to the `imslogin.test.js` test-file, which tries to achieve the same (_/nala/tests/milo/imslogin.test.js_). 

**Notable issues in the previous approach:**
1. The level of hardcoding is ***immense***! (_We should make use of page-object props for hardcoded URLS, assets, texts. When any of these URLs will change, all the test-files using them will have to be changed, instead of just the POM props._)

2. Different login types should be tested inside different test files. This current approach defeats the NALA purpose of having fast, atomic checks. Also, the `if/else` logic breaks the test-file logic into confusing chunks.  (_This logic would be much better broken down inside the POM, via `switch` statements, as that's why the structure exists._)

3. Some logic should be treated inside the POM methods, not outside of it.

**eg:**

```
        if (props.tag === '@gnav-multi-signin') {
          const multiSigninBtn = page.locator(selectors['@gnav-multi-signin']);
          await expect(multiSigninBtn).toBeVisible();
          await multiSigninBtn.click();
        }

        if (props.url.includes('bacom--adobecom')) {
          await page.waitForURL('**/auth-stg1.services.adobe.com/en_US/index.html**/');
          await expect(page).toHaveURL(/.*auth-stg1.services.adobe.com/);
        } else {
          await page.waitForURL('**/auth.services.adobe.com/en_US/index.html**/');
          await expect(page).toHaveURL(/.*auth.services.adobe.com/);
        }
 ```
 
4. The current way of writing selectors like key-value pairs inside the `/nala/selectors`, then importing them into the test-file, pollutes the context, enlarges the file footprint and greatly affects visibility (_imagine lengthier flows!_), as well as defeating the purpose of already having them written down once (***INSIDE THE POM!***).

**eg:**

```
        const ecSigninBtn = page.locator(selectors['@gnav-ec-signin']);
        await expect(ecSigninBtn).toBeVisible();
        const commSigninBtn = page.locator(selectors['@gnav-comm-signin']);
        await expect(commSigninBtn).toBeVisible();
```        

Now, lets finally compare the two for visibility and readability purposes:

```
      // Initialize FEDS login page:
      const Login = new FedsLogin(page);
      const Header = new FedsHeader(page);
      // Load FEDS user profile page:
      await page.goto(props.url);
      // Wait for page to load & stabilize:
      await page.waitForLoadState('domcontentloaded');

      // Wait for FEDS GNAV to be visible:
      await Header.MainNavLogo.waitFor({state: 'visible', timeout: 10000});
      await Header.MainNavContainer.waitFor({state: 'visible', timeout: 10000});

      // Login with a valid ACOM account:
      await Header.SignInLabel.click();
      await Login.loginOnAppForm(process.env.IMS_EMAIL, process.env.IMS_PASS);

      // Check FEDS user profile:
      await Header.openUserProfile();
      await expect(Header.ProfileModal).toBeVisible();
      await Header.checkUserProfile();
      await Header.closeUserProfile();
      await expect(Header.ProfileModal).not.toBeVisible();
      await Header.logoutUser();
      
```

**vs.**

```
        await page.goto(props.url);

        // Sign-in
        await ims.clickSignin(page);

        if (props.tag === '@gnav-multi-signin') {
          const multiSigninBtn = page.locator(selectors['@gnav-multi-signin']);
          await expect(multiSigninBtn).toBeVisible();
          await multiSigninBtn.click();
        }

        if (props.url.includes('bacom--adobecom')) {
          await page.waitForURL('**/auth-stg1.services.adobe.com/en_US/index.html**/');
          await expect(page).toHaveURL(/.*auth-stg1.services.adobe.com/);
        } else {
          await page.waitForURL('**/auth.services.adobe.com/en_US/index.html**/');
          await expect(page).toHaveURL(/.*auth.services.adobe.com/);
        }

        await ims.fillOutSignInForm(props, page);

        if (props.tag === '@gnav-multi-signin') {
          // Open App Launcher
          const appLauncher = page.locator(selectors['@gnav-app-launcher']);
          await expect(appLauncher).toBeVisible();
          await appLauncher.click();

          // Verify the list of apps is available and all are showing.
          const appsList = await page.locator(selectors['@app-launcher-list']).count();
          expect(appsList).toEqual(12);

          // Verify the apps can be clicked and navigated too.
          const ccApp = page.locator(selectors['@cc-app-launcher']);
          await expect(ccApp).toBeVisible();

          const [newPage] = await Promise.all([
            context.waitForEvent('page'),
            await ccApp.click(), // Opens a new tab
          ]);
          await newPage.waitForLoadState();
          await newPage.waitForURL('https://creativecloud.adobe.com/?context=home_cc&locale=en');
          await expect(newPage).toHaveURL(/.*creativecloud.adobe.com\/\?context=home_cc&locale=en/);
        }

        // Sign-out
        await page.locator(selectors['@gnav-profile-button']).click();
        const viewAccount = page.locator(selectors['@gnav-viewaccount']);
        expect(viewAccount).toBeVisible();
        const signoutBtn = page.locator(selectors['@gnav-signout']);
        expect(signoutBtn).toBeVisible();
        await signoutBtn.click();
        await page.waitForURL(`${props.url}#`);
        expect(page).toHaveURL(`${props.url}#`);

        const signinBtn = page.locator(selectors['@gnav-signin']);
        await expect(signinBtn).toBeVisible();
      });
    }

    if (props.tag === '@gnav-multi-signin') {
      test(`${props.title} (Drop-Down Options Check)`, async ({ page }) => {
        await page.goto(props.url);

        // Sign-in
        await ims.clickSignin(page);
        const ecSigninBtn = page.locator(selectors['@gnav-ec-signin']);
        await expect(ecSigninBtn).toBeVisible();
        const commSigninBtn = page.locator(selectors['@gnav-comm-signin']);
        await expect(commSigninBtn).toBeVisible();
        await ecSigninBtn.click();
        await page.waitForURL(/.*auth.services.adobe.com\/en_US.*/);
        await expect(page).toHaveURL(/.*auth.services.adobe.com/);
      });
    }
```

### CODE REVIEW

File: pages/feds.login.page.js [hunk 1]
```diff
@@ -0,0 +1,93 @@
...
+
+  async loginOnAppForm(email, password) {
+    console.info(`[EuroLogin] APP login form identified!`);
+    console.info(`[EuroLogin] Logging in with '${email}' account ...`);
+    // Wait for page to load & stabilize:
+    await this.page.waitForLoadState('domcontentloaded');
+    // Wait for the SUSI login form to load:
+    await this.AppEmailForm.waitFor({state: 'visible', timeout: 10000});
+    // Insert account email & click 'Continue':
+    await this.AppEmailField.waitFor({state: 'visible', timeout: 10000});
```
skumar09, line 44: - Do we need these `waitFor({state: 'visible', timeout: 10000})` twice
- can these statement/step waits can be configured from `playwright.config.js`

skumar09 (commented):
- can we use existing `<selectors>` folder for rather having separate folder `<pages>`
- and naming as `<fed.head.selector.js>` than `<fed.head.page.js>`

This is my personal view on naming convention, I am open for others view...
@JackySun9  @amauch-adobe  any thoughts...

pages/feds.login.page.js
amauch-adobe: there certainly can be added defaults for timeouts, but some things may require more time than the default.  If that is the case, it would make sense to add it for those one-off situations.  The default timeout for test.config calls are 30 sec, for expect calls it is 5 sec.

amauch-adobe:
> * can we use existing `<selectors>` folder for rather having separate folder `<pages>`
> * and naming as `<fed.head.selector.js>` than `<fed.head.page.js>`
> 
> This is my personal view on naming convention for uniformity with respect to current structure, I am open for others view... @JackySun9 @amauch-adobe any thoughts...

In my opinion, the naming conventions for fed.head.page.js seem right since it is more than just selectors in the file and could or should have methods, etc., and not just selectors.  Playwright's example class uses Page in their name so I think it makes sense to also keep the page.js in the naming construct.  We are doing something similar with the Visual Tests having the vc.test.js (visual comparison test) as its nomenclature.

I know we are trying to keep things slim and minimalistic for the Nala project, but in this case, I think it makes sense to have a pages directory instead of having POMs be housed in the selector's directory since as stated above POMs are more than just a file of selectors, but also methods, etc.

cc @skumar09 @JackySun9 @chadsunvice

pages/feds.header.page.js
```diff
@@ -0,0 +1,64 @@
...
+    this.ProfileSignOut = page.locator('a.feds-profile-action');
+
+    // GNAV breadcrumb selectors:
+    this.BreadcrumbList = page.locator('nav.feds-breadcrumbs');
+    this.BreadcrumbElems = page.locator('nav.feds-breadcrumbs li');
+    this.BreadcrumbContainer = page.locator('div.feds-breadcrumbs-wrapper');
+  }
+
+  async openUserProfile() {
+    await this.ProfileIcon.waitFor({state: 'visible', timeout: 10000});
```
amauch-adobe, line 35: if there is a reason for adding the timeout then let's consider it, otherwise like I mentioned in my reply to Santosh's comment.  The default for test.config method calls (which I think .waitFor fits under) is 30 seconds globally.  expect() method calls default at 5 seconds globally.  I do believe though, increased timeouts can be added if needed beyond the defaults.

pages/feds.login.page.js
amauch-adobe: oh, that being said though, the Login page is a third-party service and not a Milo/Franklin-based page and required increased timeouts for various checks/waits since the form takes longer to load than general Milo/Franklin blocks.  Plus, the Login form has to switch its content when you submit the IMS email account to the IMS password field they both don't just show up at the start.  This is why I believe Daniel might be having visible checks with timeouts.

pages/feds.login.page.js
```diff
@@ -0,0 +1,93 @@
...
+    this.PersonalAccountLogo = page.locator('img[alt="Personal Account"]');
+    this.SelectAccountForm = page.locator('div[data-id="Profile"]');
+
+    this.AppEmailFieldSelector = page.locator('input#EmailPage-EmailField');
+    this.AppPasswordFieldSelector = page.locator('input#PasswordPage-PasswordField');
+    this.CodePadChallenge = page.locator('div[data-id="ChallengeCodePage"]');
+  }
+
+  async loginOnAppForm(email, password) {
+    console.info(`[EuroLogin] APP login form identified!`);
```
amauch-adobe, line 37: Just want to have some clarification for the console logging.  Playwright says to only use console logging for debugging purposes, but if there is a valid reason to have it, then I say go for it.

pages/feds.login.page.js
```diff
@@ -0,0 +1,93 @@
...
+    await expect(this.AppPasswordContinue).toHaveText('Continue');
+    await this.AppPasswordContinue.click();
+    // Insert account password & click 'Continue':
+    await this.AppPasswordForm.waitFor({state: 'visible', timeout: 10000});
+    await this.AppPasswordField.waitFor({state: 'visible', timeout: 10000});
+    await this.AppPasswordField.fill(password);
+    await this.AppLoginContinue.waitFor({state: 'visible', timeout: 10000});
+    await expect(this.AppLoginContinue).toHaveText('Continue');
+    await this.AppLoginContinue.click();
+    // Assert the login process was successful:
```
amauch-adobe, line 56: Silly nitpick but when you used the word Assert I first was thinking that maybe there would be an expect() method coming instead of just waiting for something to become visible. Don't worry about it. ;). I'm just thinking out loud about what I first thought reading your comment.

pages/feds.login.page.js
```diff
@@ -0,0 +1,93 @@
...
+    this.AppPasswordContinue = page.locator('button[data-id^="EmailPage"]');
+    this.AppLoginContinue = page.locator('button[data-id^="PasswordPage"]');
+    this.PersonalAccountLogo = page.locator('img[alt="Personal Account"]');
+    this.SelectAccountForm = page.locator('div[data-id="Profile"]');
+
+    this.AppEmailFieldSelector = page.locator('input#EmailPage-EmailField');
+    this.AppPasswordFieldSelector = page.locator('input#PasswordPage-PasswordField');
+    this.CodePadChallenge = page.locator('div[data-id="ChallengeCodePage"]');
+  }
+
```
amauch-adobe, line 35: I think it would be provincial to add a block comment providing a method definition as a standard for POM methods. That way when people call the method while developing in an IDE they can see the method description/definition, params, etc.

pages/feds.login.page.js
```diff
@@ -0,0 +1,93 @@
...
+    await this.PasswordField.waitFor({state: 'visible', timeout: 10000});
+    await this.PasswordField.fill(password);
+    // Complete the login flow:
+    await this.LoginButton.waitFor({state: 'visible', timeout: 10000});
+    await this.LoginButton.click();
+    // Assert the login process was successful:
+    await this.LoggedInState.waitFor({state: 'visible', timeout: 10000});
+    console.info(`[EuroLogin] Successfully logged-in as '${email}' (via SUSI login form).`);
+  }
+
```
amauch-adobe, line 85: Might be worth providing a warning on using this method to not have a valid password be used for this method since it obviously will show it up on the VMs.  I'm not sure how big a security risk it is since it is so short of a method check, but it never hurts. :)

amauch-adobe (changes requested):
Love how clean the test is, just had some comments/questions about using console logging where Playwright's docs state to only use it for debugging.  If there is a valid reason for the logging, then let's do it.  My other larger request for changes is adding method definition block comments for each POM method that way usage in IDEs will be easier since it will bring up the definition as well as what type of parameter(s) need to be sent in.  See my other comments throughout.

pages/feds.login.page.js
chadsunvice: > * Do we need these `waitFor({state: 'visible', timeout: 10000})` twice
> * can these statement/step waits can be configured from `playwright.config.js`

This is a good spot @skumar09, thanks! I hastily prepared the `loginOnSusiForm` method for checks where the `#App` version is not loaded. We can get to test it in MAX/Summit scenarios in the future and polish it when needed. I updated the method.

chadsunvice: > there certainly can be added defaults for timeouts, but some things may require more time than the default. If that is the case, it would make sense to add it for those one-off situations. The default timeout for test.config calls are 30 sec, for expect calls it is 5 sec.

Yes, I totally agree @amauch-adobe! The problem with the login checks is that it takes a bit of time to load everything and we have no control over when IMS returns the success message for any call. I've used `10000ms` as the default for Login. In other scenarios, I've seen it is not needed. In the future we can custom tailor these in the `playwright.conf.js` file. I'm still learning what timeouts work best and adapting the scripts as we go.

chadsunvice: > oh, that being said though, the Login page is a third-party service and not a Milo/Franklin-based page and required increased timeouts for various checks/waits since the form takes longer to load than general Milo/Franklin blocks. Plus, the Login form has to switch its content when you submit the IMS email account to the IMS password field they both don't just show up at the start. This is why I believe Daniel might be having visible checks with timeouts.

Exactly! I've seen `10000ms` to be the sweet-stop for when dealing with IMS requests. I ran the checks multiple times before sending them, and Firefox/Chromium always finishes fast, but on WebKit, there would always be failures when `5000ms` wait time was used. But, since this is not an explicit wait, it is no issue to have a bigger wait time, just a safety net to avoid brittle checks and unexpected failures.

pages/feds.login.page.js
chadsunvice: This is usually flavour imho and I don't have a strong position on this. I am open to changing it as NALA team seems fit. 

For the last 12y I've dealt with multiple testing frameworks and 100% I preferred the checks with more logging than not. I see them as success flags to track the execution of steps in case of debugging, when the issue is not easily spottable via stack-trace errors.

I suggest to either remove them altogether, or push them with a logger to a debug file of sorts. I am open to suggestions.

pages/feds.login.page.js
chadsunvice: That's a great idea. I agree with this! I'll look into it and update soon.

pages/feds.header.page.js
chadsunvice: > > * can we use existing `<selectors>` folder for rather having separate folder `<pages>`
> > * and naming as `<fed.head.selector.js>` than `<fed.head.page.js>`
> > 
> > This is my personal view on naming convention for uniformity with respect to current structure, I am open for others view... @JackySun9 @amauch-adobe any thoughts...
> 
> In my opinion, the naming conventions for fed.head.page.js seem right since it is more than just selectors in the file and could or should have methods, etc., and not just selectors. Playwright's example class uses Page in their name so I think it makes sense to also keep the page.js in the naming construct. We are doing something similar with the Visual Tests having the vc.test.js (visual comparison test) as its nomenclature.
> 
> I know we are trying to keep things slim and minimalistic for the Nala project, but in this case, I think it makes sense to have a pages directory instead of having POMs be housed in the selector's directory since as stated above POMs are more than just a file of selectors, but also methods, etc.
> 
> cc @skumar09 @JackySun9 @chadsunvice

I totally agree @amauch-adobe!

pages/feds.login.page.js
chadsunvice: > there certainly can be added defaults for timeouts, but some things may require more time than the default. If that is the case, it would make sense to add it for those one-off situations. The default timeout for test.config calls are 30 sec, for expect calls it is 5 sec.

From my testing, the `waitFor` method doesn't share the global `30000ms` timeout you mentioned @amauch-adobe . I didn't read the comment well at first. If I remove the timeouts, the checks will become flaky, or fail altogether. I kept `10000ms` for consistency throughout those methods. As I said, it is not explicit waiting. There's polling done behind the scenes.

pages/feds.login.page.js
amauch-adobe: lets keep them for now, and as we look into better logging options we can adjust.

pages/feds.header.page.js
amauch-adobe: Does it use the 30-sec timeout in the config if their timeout option isn't applied to the waitFor method?  If it doesn't then we should probably log an issue with Playwright's project in GitHub:  https://github.com/microsoft/playwright/issues

I'm good for keeping your options in, I believe the global configuration is the stopgap, but I'm okay when developers add their own timeout option as needed.

amauch-adobe (approved):
Approved.

skumar09:
@chadsunvice @amauch-adobe 
- one small input: can we use camelCase for identifier names (variables and functions), [ I think milo code base uses camelCase naming convention ]

pages/feds.login.page.js
skumar09: @chadsunvice @amauch-adobe 
- You may have console.info(), but I would not recommend logging user or profile information like email in GitHub

amauch-adobe:
@chadsunvice I added a new label for associating your tags with executions in Nala, i.e. @feds-userprofile.  When Nala ran against your PR it looks like 3 test runs failed for the user profile test.  Looking at the exception it is pointing to the waitFor timeout value.  I think you were right in that 5 seconds is too small and needs to be updated back to your 10-second timeout.

Screenshot 2023-03-30 at 9 58 25 AM

chadsunvice:
> @chadsunvice I added a new label for associating your tags with executions in Nala, i.e. @feds-userprofile. When Nala ran against your PR it looks like 3 test runs failed for the user profile test. Looking at the exception it is pointing to the waitFor timeout value. I think you were right in that 5 seconds is too small and needs to be updated back to your 10-second timeout.
> 
> Screenshot 2023-03-30 at 9 58 25 AM

Roger that Aaron! @amauch-adobe . I'll update scripts and push again. Hopefully this time it passes the checks! IMS is often unreliable, it can go fast, but it can lag sometimes.

amauch-adobe (approved):
Approved.

amauch-adobe:
hmm, maybe try 15 seconds I know the old IMS test I made required 15 seconds for the call that keeps failing below for all of the browsers for a particular form switch.

chadsunvice:
> hmm, maybe try 15 seconds I know the old IMS test I made required 15 seconds for the call that keeps failing below for all of the browsers for a particular form switch.
> 
> Screenshot 2023-03-30 at 3 43 51 PM

For now I increased only for login methods to 15000ms, as proposed.

In the future we should look into how we can make maybe some API calls for most login flows as this might be a bottleneck going forward. I'll research the subject and write a post in #milo-community when I find a more elegant approach.

We should strive to keep the checks as fast and atomic as possible as to leverage the power of parallel execution across multiple workers that PlayWright offers.

JackySun9 (approved):

chadsunvice:
@amauch-adobe, I increased the timeouts, but it seems it still fails to get the password field visible, which I could not reproduce locally. I think I will add a screenshot taking method to see the last state before test fails, cause it might be something else.

amauch-adobe:
> @amauch-adobe, I increased the timeouts, but it seems it still fails to get the password field visible, which I could not reproduce locally. I think I will add a screenshot taking method to see the last state before test fails, cause it might be something else.

I think that is a great idea, lets's go in that direction and see what we can find.  There is also a GitHub Action called [Nala Debugging](https://github.com/adobecom/nala/actions/workflows/debug.yml) that you can fire against your branch and test tags and it will supply a playwright report with a trace view you can use to look at [here](https://trace.playwright.dev/) are on your local machine later via command-line.

You can use these values to fire the GitHub action although you will need to fire it off on your fork's repo and not the link I just sent you. :)

branch: MWPW-128180_feds_user_profile
tags: @feds-consent @feds-userprofile
os: ubuntu-latest

chadsunvice:
> > @amauch-adobe, I increased the timeouts, but it seems it still fails to get the password field visible, which I could not reproduce locally. I think I will add a screenshot taking method to see the last state before test fails, cause it might be something else.
> 
> I think that is a great idea, let's go in that direction and see what we can find. There is also a GitHub Action called [Nala Debugging](https://github.com/adobecom/nala/actions/workflows/debug.yml) that you can fire against your branch and test tags and it will supply a playwright report with a trace view you can use to look at [here](https://trace.playwright.dev/) are on your local machine later via command-line.
> 
> You can use these values to fire the GitHub action although you will need to fire it off on your fork's repo and not the link I just sent you. :)
> 
> branch: MWPW-128180_feds_user_profile tags: @feds-consent @feds-userprofile os: ubuntu-latest

I tried this approach, but the trace available in the playwright-report folder did not load as expected.

amauch-adobe (approved):
Approved.

amauch-adobe:
So when I ran the exact code (I just copied it into the branch I created using the code from your fork's branch), created a PR, and ran it through the CI GitHub Actions, everything is passing just fine.  I then ran it through the Nala Debugging GitHub Actions Workflow and it also passed just fine.  Here are my screenshots for both and the playwright report. I think it was just struggling to get the secrets, anyways I'm good with merging your fork's code in.  I'll remove my PR and delete my branch.

The "fake/testing" PR is here #112

chadsunvice:
> So when I ran the exact code (I just copied it into the branch I created using the code from your fork's branch), created a PR, and ran it through the CI GitHub Actions, everything is passing just fine. I then ran it through the Nala Debugging GitHub Actions Workflow and it also passed just fine. Here are my screenshots for both and the playwright report. I think it was just struggling to get the secrets, anyways I'm good with merging your fork's code in. I'll remove my PR and delete my branch.
> 
> The "fake/testing" PR is here #112

Thanks a lot for helping me with this @amauch-adobe!

PR #91, opened 2023-03-27
Nala pom and tests enhancements

skumar09:
### Tests Structure:
1. In Current test structure 
	- we are executing all specs or testcases inside a features for-loop, with increase in specs counts it 
	  is getting difficult to add new test and tweak the logic, it is not scalable.
	- it is also quite difficult to read, debug and understand the tests, steps and flows. 
2. With this PR , provided following :
	- All tests are grouped under` test.describe()` that is something like **suite ex: carousel-block test suite.**
	- Each spec or **testcase from `.spec.js` file corresponds to a separate test**. (as below screenshot.)
	- Introduced `test.step()` under `test(`)
        - This will introduce better readability, separation of code / logic for different tests and test steps
3. Spec
	- Spec structure is simplified to array as follows, (we can add item as required for the specs)
	- Parsing Spec.js is not required.
	- Env `baseURL` is taken from` playwright.config.js. ( This part is still in enhancement )`
4. Page Objects.
	- Provided the sample for how to define selectors with reference to parenent selector or element.
	- Provided method, guideline, commenting format for page object methods.
### Features/Specs
image

### Tests
image

### POM
image

### DIFF
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -38,12 +38,13 @@ const config = {
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
-    actionTimeout: 30000,
+    actionTimeout: 60000,
     /* Base URL to use in actions like `await page.goto('/')`. */
     // baseURL: 'http://localhost:3000',
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: 'on-first-retry',
+    baseURL: 'https://main--milo--adobecom.hlx.page',
   },
 
   /* Configure projects for major browsers */
--- a/selectors/milo/carousel.block.page.js
+++ b/selectors/milo/carousel.block.page.js
@@ -1,0 +1,220 @@
+/* eslint-disable import/no-import-module-exports */
+import { expect } from '@playwright/test';
+
+exports.Carousel = class Carousel {
+  constructor(page) {
+    this.page = page;
+    // carousel types selectors
+    this.carouselContainer = page.locator('.carousel.container');
+    this.carouselLightbox = page.locator('.carousel.lightbox');
+    this.carouselContainerShow2 = page.locator('.carousel.show-2.container');
+    this.carousel = page.locator('.carousel');
+
+    // carousel selectors
+    this.slides = this.carousel.locator('.carousel-slides');
+    this.activeSlide = this.slides.locator('.section.carousel-slide.active');
+    this.slidesCount = this.slides.locator('.section.carousel-slide');
+    this.indicator = this.carousel.locator('.carousel-indicators');
+    this.indicatorCount = this.indicator.locator('.carousel-indicator');
+    this.activeIndicator = this.indicator.locator('.carousel-indicator.active');
+    this.nextButton = this.carousel.locator('.carousel-next');
+    this.previousButton = this.carousel.locator('.carousel-previous');
+    this.lightboxExpandButton = this.carouselLightbox.locator('.lightbox-button.carousel-expand');
+    this.lightboxCloseButton = this.carouselLightbox.locator('.lightbox-button.carousel-close');
+  }
+
+  /**
+ * Get the index of the current slide.
+ * @return {Promise<number>}.
+ */
+  async getCurrentSlideIndex() {
+    const currentIndex = await this.activeSlide.getAttribute('data-index');
+    return currentIndex;
+  }
+
+  /**
+ * Get the count of slides of carousel.
+ * @return {Promise<number>}.
+ */
+  async getNumberOfSlides() {
+    const numberOfSlides = await this.slidesCount.count();
+    return numberOfSlides;
+  }
+
+  /**
+ * Move to next slide .
+ */
+  async moveToNextSlide() {
+    await this.nextButton.click();
+  }
+
+  /**
+ * Move to previous slide .
+ */
+  async moveToPreviousSlide() {
+    await this.previousButton.click();
+  }
+
+  /**
+ * Move to nth  slide .
+ */
+  async moveToSlide(index) {
+    await this.indicator.nth(index).click();
+  }
+
+  /**
+ * Are carousel indictors are displayed.
+ * @return {Promise<boolean>}.
+ */
+  async areIndicatorsDisplayed() {
+    const isDisplayed = await this.indicator.isVisible();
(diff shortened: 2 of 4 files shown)


playwright.config.js
```diff
@@ -38,12 +38,13 @@ const config = {
...
   use: {
     /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
-    actionTimeout: 30000,
+    actionTimeout: 60000,
     /* Base URL to use in actions like `await page.goto('/')`. */
     // baseURL: 'http://localhost:3000',
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: 'on-first-retry',
+    baseURL: 'https://main--milo--adobecom.hlx.page',
```
JackySun9, line 47: Maybe we can introduce profile concept, because the baseUrl is totally different per team

selectors/milo/carousel.block.page.js
```diff
@@ -0,0 +1,220 @@
+/* eslint-disable import/no-import-module-exports */
+import { expect } from '@playwright/test';
+
+exports.Carousel = class Carousel {
```
JackySun9, line 4: POM is good, we might have some base pages, the purpose for this design is inheritance and reuse

JackySun9 (changes requested):

playwright.config.js
skumar09: @JackySun9 : this one still work in progress,[ in the playwright world we can have multiple` config.js `or `globalsetup.js` for environment, working on...]

JackySun9 (approved):

JingleH (approved):
looks good to me

selectors/milo/carousel.block.page.js
```diff
@@ -0,0 +1,220 @@
...
+ * Click carousel <lightbox expand> button
+ */
+  async expandLighboxModal() {
+    await this.lightboxExpandButton.click();
+  }
+
+  /**
+ * Click carousel <lightbox modal close> button
+ */
+  async closeLighboxModal() {
```
rgclayton, line 145: Typo. Should be `closeLightboxModal`

tests/milo/carousel.block.test.js
```diff
@@ -0,0 +1,124 @@
...
+
+      expect(await carousel.isNextButtonlVisible()).toBeTruthy();
+      expect(await carousel.isPreviousButtonlVisible()).toBeTruthy();
+
+      // verify expand and close lightbox
+      expect(await carousel.isLightboxExpandButtonVisible()).toBeTruthy();
+      await carousel.expandLighboxModal();
+
+      expect(await carousel.isLightboxCloseButtonVisible()).toBeTruthy();
+      await carousel.closeLighboxModal();
```
rgclayton, line 88: Typo. Should be `closeLightboxModal`

rgclayton (commented):
Found a minor typo

playwright.config.js
amauch-adobe: If I'm considering this right, when Nala becomes a consumable tool teams should be able to configure the baseURL to their desired one for their team.  Having so many profiles/configs will get bulky in my opinion and for core Nala, if it isn't going to be configurable I think this will become bulky and heavy on maintenance.

tests/milo/carousel.block.test.js
```diff
@@ -0,0 +1,124 @@
...
+import { Carousel } from '../../selectors/milo/carousel.block.page';
+
+const CarouselSpec = require('../../features/milo/carousel.block.spec');
+
+const { features } = CarouselSpec;
+
+// Carousel blocks tests
+test.describe('Milo Carousel block test suite', () => {
+  // Test - 1
+  test(`${features[0].name}, @milo-live, ${features[0].tags},https://milo.adobe.com `, async ({ page, baseURL }) => {
```
amauch-adobe, line 13: where the envs is being added as @milo_live and @milo_prod having "https://milo.adobe.com added to the test title would make it incorrect since @milo_live is actually https://main--milo--adobecom.hlx.live.  So the resulting report would provide the incorrect URL it was tested against.  Also if you are adding envs as a part of the spec for environments those paths should be tested against, how does the current test setup work to test against those different environments?  Currently, from what I can see this will only test against what the base URL is which may not ever be any of those environments in the spec file.  As mentioned in my previous comment the parse.js lib was setup so that you could test a block against many different environments for the same path since you could run a test against hlx.page or hlx.live or a production URL with the same path.  I guess looking at the current commit for the spec, we would have to add a test() for every feature spec array we make and only have one env variable for them even though the tags and paths would be the same, that is one way of getting around this problem but it also adds a bit of redundancy in feature spec.

utils/extract-test-title.js
amauch-adobe, line 1: I'm okay with the adjustments, I just want to ensure that things are still grouped correctly in the Nala Dashboard App for test results.  I also want to ensure the filtering will still work.  @JingleH do we know if that is the case?

amauch-adobe (changes requested):
I like adding the direction for more structured and readable test scripts by using the out-of-the-box playwright annotations, test.step() methods, etc.  I think that is a great adjustment and move, I think having the baseURL configurable is the best option instead of creating more complex profiles and bulking up Nala with tons of different config files per team.  I also think the baseURL is a default option to use if it works for your test, but shouldn't be always used (see my comments about PR testing, the usage of parse.js with feature branch testing, and the ?milolibs= URL parameter).  I think the tcid addition to specs will also help with the TestRail integration.  The POM model looks great, and fits exactly what we have already discussed with Daniel Chivesco with his PRs.  I also don't want us to forget the other mission of Nala was to make it simple for all developer types and not just QEs or testers.  One of the reasons for creating Nala was to bridge gaps between all team members to feel comfortable in using a tool to create End-to-End tests and to do so quickly.  Whenever we add, change, or switch to a new way of doing things we need to make sure there is buy-in to the changes, don't remove key functionality that is core to Nala, and keep things simple for all developers.  @skumar09, I think having a huddle on this one would be good for me so I can be on the same page for some of the changes.

utils/extract-test-title.js
JingleH: I was being lazy and simply hoping to see if this breaks the dashboard or not haha. If we decide to do things this way and merge this PR, I can make corresponding changes in the dashboard if needed. Looks like it's sending all the required params, so hopefully nothing breaks. Thanks for letting me know!

amauch-adobe (approved):
Approved, had huddle with @skumar09 everything looks good, and answered concerns.

rgclayton (approved):

PR #92, opened 2023-03-28
DC converter and review block selector updates

JFernandezAdobe:
1. Update the converter selector to fix the failure in Compress PDF test after the widget upgrade on 3/28.
2. Simplify the selectors used in the review block test


skumar09:
@JFernandezAdobe: Few tests are failing with WebKit browser, can you please review, [ not sure if it is a known issue or if it expected ]

selectors/dc/dc_converter.selectors.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
   '@excel-file-upload-input': '[data-test-id="file-upload-input"][accept*=".xlsx"]',
   '@export-convert-button': '[data-test-id="export-form-convert-button"]',
   '@export-progress-bar': '.spectrum-BarLoader-label',
-  '@convert-button': '[data-test-id="convert"]',
+  '@convert-button': '[data-test-id="convert"], [data-test-id*=compress-button]',
```
amauch-adobe, line 10: Does this work with having two different selectors for one variable?  Just curious since they are on the same line but are comma-separated, so I figured it would break.

amauch-adobe (changes requested):
Does the updated selector variable that looks to be assigned two different selectors (separated by a comma on one line) work?  I would think it would fail since I assume it would expect the comma-delimited second selector to be assigned a different variable.

selectors/dc/dc_converter.selectors.js
JFernandezAdobe: Yes, it works. The tests that use the same selector passed.

amauch-adobe: sounds good.

amauch-adobe:
I'm okay approving once all tests pass, it seems like the webkit tests for DC has been difficult to get to pass consistently.  What is your take on it @JFernandezAdobe?

JFernandezAdobe:
@amauch-adobe @skumar09 I can't reproduce the test issues in Webkit locally. Skipping Webkit for now until there is a solution in place.

amauch-adobe (approved):
Approved.

chrischrischris (approved):

skumar09 (approved):

PR #94, opened 2023-03-29
'run-nala' label update to 'run-nala-<application>' label

amauch-adobe:
This PR resolves this Jira Ticket: [MWPW-128629](https://jira.corp.adobe.com/browse/MWPW-128629)

- moved review spec
- added @milo tag
- Updated Github actions and run.sh to use new regex label run-nala-<application> so the correct application tag can be added to playwright execution.


amauch-adobe:
@skumar09 is taking point on this ticket now: [MWPW-128629](https://jira.corp.adobe.com/browse/MWPW-128629)

PR #96, opened 2023-03-30
[MWPW-128064] Header/Footer component checks

chadsunvice:
* Header/Footer infrastructure
* Header first checks (a lot of the logic is under development on Milo side so these will be updated)

### DIFF
--- a/pages/feds.header.page.js
+++ b/pages/feds.header.page.js
@@ -1,0 +1,95 @@
+import { expect } from '@playwright/test';
+
+exports.FedsHeader = class FedsHeader {
+
+  constructor(page) {
+    this.page = page;
+
+    // GNAV selectors:
+    this.GnavLogo = page.locator('a.gnav-logo');
+    this.MainNavLogo = page.locator('a.feds-brand');
+    this.MainNavContainer = page.locator('nav.feds-topnav');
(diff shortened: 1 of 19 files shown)


tests/feds/footer.test.js
amauch-adobe, line 1: Perchance is this supposed to be empty? ;)

pages/feds.header.page.js
amauch-adobe, line 1: Looks like this file is also in another PR #90, just want to make sure they stay in sync.

amauch-adobe (changes requested):
Footer.test.js is an empty file, don't know if it should be committed empty.  I also added the new tag labels you added in the two new spec files and added Nala to run against your PR.

pages/feds.header.page.js
chadsunvice: I've synced them beforehand to have no issue when merging. I have the latest version into the 3 PRs open.

tests/feds/footer.test.js
chadsunvice: I can remove the file for now, I agree. The Footer isn't ready yet and this was a placeholder of sorts.

amauch-adobe (approved):
Approved.

JackySun9 (approved):

PR #98, opened 2023-04-04
Include DC pages in columns and feds tests

JFernandezAdobe:
Run the existing columns and feds tests in Frictionless pages.


JFernandezAdobe:
https://github.com/adobecom/nala/actions/runs/4620467302/jobs/8170610890?pr=98
`Notice:   69 passed (3.1m)`

features/milo/failedblock.spec.js
amauch-adobe, line 1: Are you guys allowing the failedblock code to run on production URLs?  As far as I know, the current code configuration for the FailedBlock is to only show up on hlx.page or hlx.live URLs, this is by design so public consumers won't see the failed block wrapper.  On production, if a failed block is present it will just show up as empty whitespace on the page.  Are you guys doing something special with your setup to allow the failedblock wrapper to show up on production?

pages/feds.header.page.js
amauch-adobe, line 1: @chadsunvice I think it would be good for you to look over this PR since there are updates being made to the original specs and poms for the feds pages.

amauch-adobe (changes requested):
See my comments on the hlx.live url dc domain change to the failedblock test.

features/milo/failedblock.spec.js
JFernandezAdobe: > As far as I know, the current code configuration for the FailedBlock is to only show up on hlx.page or hlx.live URLs, this is by design so public consumers won't see the failed block wrapper.

@amauch-adobe I have seen the failed block message before in stage.adobe.com, not only in hlx.page or hlx.live URLs. Please see [MWPW-125609](https://jira.corp.adobe.com/browse/MWPW-125609).

@Blainegunn could you confirm if this is not the case anymore? Should we create a ticket to investigate this?

Also, the reason for not running DC tests in hlx.live and replacing them with prod URLs is same as https://github.com/adobecom/nala/pull/77.

amauch-adobe: @JFernandezAdobe I can understand that stage.adobe.com shows it up since you guys are filtering hlx.page to that domain, I just didn't think the failed block would show up for www.adobe.com production URLs since that is public-facing and shouldn't show the wrapper.  I was just saying we needed to pick a URL that would work for the wrapper to show up on.

amauch-adobe: @JFernandezAdobe I'm okay with approving this PR, I would like to make sure whether or not the failedblock wrapper shows up on the UI for www.adobe.com URLs since that is production public facing URLs and per the failed block design it shouldn't be able to.  Also, I really would like @chadsunvice to first have a look for the FEDS test side of things as well.

amauch-adobe (approved):
Approved, pending my question about the failed block wrapper showing up against www.adobe.com and @chadsunvice having a review of the FEDS test side of things.

PR #101, opened 2023-04-09
MWPW-129400 update consumer config and baseURL structure

skumar09:
[MWPW-129400](https://jira.corp.adobe.com/browse/MWPW-129400) : Enhance Nala framework playwright.config.js projects and baseURL structure.

This PR fixes the following current issues:

1. Test scripts are executed as per the value(s) specified in the `env` param in `.spec.js files`.  
    Ex: In the below example spec, test script execution is repeated against each tag and each environment. i.e. in a single GitHub run, the below script is run four times, which is not required. Also, the user should be able to run the tests against specified tags in any given environments.
```
  features: [
    {
      tcid: '1',
      name: '@Carousel(container)',
      path: '/drafts/nala/blocks/carousel/lightbox',
      tags: '@carousel @carousel-container @smoke @regression @milo,',
      envs: '@milo-live @milo-prod',
    },]

```
2. Test execution on multiple environments or URLs is tightly coupled with `.spec.js files`.
    - Ideally, the user should be able to run scripts against any given environment, (provided the feature is available in that environment)
    - User should be able to instruct from GitHub (or outside the `.spec.js `or test cases files ) to run the test scripts on a specified environment.    
3. Consumer does not have separate config.js files to maintain their own sub-projects and URLs for their run.

**Updates about PR:**
1. 
image
2. 
image
3.
image

PR #104, opened 2023-04-14
MWPW-128629 : Update github self.yml job run check condition

skumar09:
Currently, Nala tests are run if PR has the label: `run-nala`, this is now enhanced to below

Jira : [MWPW-128629](https://jira.corp.adobe.com/browse/MWPW-128629)

User will able to run Nala tests if PR has the any label(s) in below format

- No need to have a separate label: `nala-run` to trigger the test run job
-  PR labels 
     - labels = `run-dc` or` run-nala` or `run-milo` etc  will trigger the test run job and executes tests corresponding all projects specified in config.js
     - labels = `run-milo-live` or `run-nala-milo-live` or `run-dc-stage` or `run-dc-live` etc  will trigger the test run job and executes tests corresponding to projects [ like live, stage etc] of specified in config.js
     - labels = `run-milo-live`, `@carousel-container`, `@smoke`  will trigger the test run job and executes tests corresponding to the given tags for given projects [ like live, stage etc] that are specified in config.js

This PR is a continuation to the changes done as part of [PR-101](https://github.com/adobecom/nala/pull/101)

### DIFF
--- a/.github/workflows/self.yml
+++ b/.github/workflows/self.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   action:
     name: Running tests
-    if: github.event.pull_request.labels.some(label => label.name.startsWith('run-'))
+    if: ${{ contains(join(github.event.pull_request.labels.*.name, ' '), 'run-') }}
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


JackySun9 (approved):

amauch-adobe (approved):
Approved.

PR #106, opened 2023-04-17
Adding stats block test.

Dli3:
Adding tests for the stats block on bacom.

Ticket: https://jira.corp.adobe.com/browse/MWPW-126249

### DIFF
--- a/selectors/bacom/stats.page.js
+++ b/selectors/bacom/stats.page.js
@@ -1,0 +1,21 @@
+exports.Stats = class Stats {
+  constructor(page) {
+    this.page = page;
+    this.statsLink = page.locator('.stat > div > p > strong > a');
+    this.solutionsLink = page.locator('.solution > div > p:last-of-type > a');
+  }
+
+  /**
+   * Selects the stats link of the stats block.
+   */
+  async clickStatsLink() {
+    await this.statsLink.click();
+  }
+
+  /**
+   * Selects the solutions link of the stats block.
+   */
+  async clickSolutionsLink() {
+    await this.solutionsLink.click();
+  }
+};
--- a/tests/bacom/stat.test.js
+++ b/tests/bacom/stat.test.js
@@ -1,0 +1,30 @@
+import { expect, test } from '@playwright/test';
+import { Stats } from '../../selectors/bacom/stats.page.js';
+
+const StatsSpec = require('../../features/bacom/stats.spec.js');
+
+const { features } = StatsSpec;
+
+test.describe('BACOM Stats block test suite', () => {
+  test(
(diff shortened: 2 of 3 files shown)


JackySun9 (approved):

selectors/bacom/stats.page.js
```diff
@@ -0,0 +1,21 @@
+exports.Stats = class Stats {
+  constructor(page) {
+    this.page = page;
+    this.statsLink = page.locator('.stat > div > p > strong > a');
```
amauch-adobe, line 4: Just a quick question does the final locator of the links have a unique class name or id?  If they do it is recommended by Playwright to use those before using CSS paths.  If not, no worries.

tests/bacom/stat.test.js
```diff
@@ -0,0 +1,32 @@
...
+const { features } = StatsSpec;
+
+test.describe('BACOM Stats block test suite', () => {
+  test(
+    `${features[0].name}, @bacom-live, ${features[0].tags}, https://bacom.adobe.com`,
+    async ({ page, baseURL }) => {
+      const stats = new Stats(page);
+      const testPage = `${baseURL}${features[0].path}`;
+
+      await page.goto(testPage);
```
amauch-adobe, line 15: I would just add these two lines (15-16) into the test step method since you do these same calls again just before starting the next test step, then it is all tied to the test step.

tests/bacom/stat.test.js
```diff
@@ -0,0 +1,32 @@
...
+      await page.waitForLoadState('domcontentloaded');
+
+      await test.step('Selecting a stats link', async () => {
+        await stats.clickStatsLink();
+        await expect(page.url()).not.toBe(testPage);
+      });
+
+      await page.goto(testPage);
```
amauch-adobe, line 23: Same here as above.  I would just add these two lines (23-24) into the test step method then it is all tied to the test step.

amauch-adobe (changes requested):
Great job on POM usage and test step methods, see my comments inline.

selectors/bacom/stats.page.js
Dli3: They don't. But thanks for bringing that up!

tests/bacom/stat.test.js
Dli3: Great suggestion. The lines have been moved. Thanks!

tests/bacom/stat.test.js
Dli3: Great suggestion. The lines have been moved. Thanks!

skumar09 (approved):

amauch-adobe (approved):
Approved.

skumar09:
@Dli3: Earlier it seems an issue with the labels, I have updated the labels, and the tests are passed, please review once before merging.

PR #113, opened 2023-05-04
MWPW-130636 : Update test folder path in configs

skumar09:
[MWPW-130636](https://jira.corp.adobe.com/browse/MWPW-130636) Update the test folder path in the `config.js` files for all consumer projects
PR fixes the below issue: 

- **Issue**:  If the PR Label has a project i.e. `run-milo-live` and does not have any `tags i.e. @marquee or @howto`, then all nala tests are running.

- **Expected**: If the PR Label has a project and no tags, then it should execute the test corresponding to the project.

### DIFF
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -12,7 +12,7 @@ const { devices } = require('@playwright/test');
  * @type {import('@playwright/test').PlaywrightTestConfig}
  */
 const config = {
-  testDir: './tests',
+  testDir: './tests/milo',
   outputDir: './test-results',
   /* Maximum time one test can run for. */
   timeout: 30 * 1000,
--- a/configs/acom.config.js
+++ b/configs/acom.config.js
@@ -8,7 +8,7 @@ const envs = require('../envs/envs.js');
  * @type {import('@playwright/test').PlaywrightTestConfig}
  */
 const config = {
-  testDir: '../tests',
+  testDir: '../tests/acom',
   outputDir: '../test-results',
   /* Maximum time one test can run for. */
   timeout: 30 * 1000,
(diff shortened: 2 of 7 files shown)


JackySun9 (approved):

playwright.config.js
```diff
@@ -12,7 +12,7 @@ const { devices } = require('@playwright/test');
  * @type {import('@playwright/test').PlaywrightTestConfig}
  */
 const config = {
-  testDir: './tests',
+  testDir: './tests/milo',
```
amauch-adobe, line 15: quick question, should this stay as ./tests instead of ./tests/milo since it is the generic playwright.config.js? Where we already have a milo.config.js that has the same testDir value I just was wondering if the duplication is necessary.  
I can see the value of having it in the generic one, but if we make this change now, the daily runs in GitHub will only test Milo's tests and not every other team's tests.  I know Chris M. mentioned to everyone that until we have a project set up entirely for other teams to consume and run in their own CI/CD pipelines the Nala project would run each team's tests for the time being.

amauch-adobe (changes requested):
Everything looks good to me, just had one concern.  See my comments.

playwright.config.js
skumar09: @amauch-adobe 
- currently no, because in default ` config.js`, projects specified are of milo, so we should change it to `.tests/milo` 
- In parallel, I am working on another issue that @Dli3 has discussed, where we will be handling executing other tests also.

amauch-adobe: okay, my biggest concern was for the daily runs anyways, so if things are being taken care of for other team's tests.  I approve of the change. 👍🏼

amauch-adobe (approved):
Approved.

PR #118, opened 2023-05-10
[MWPW-130049] FEDS Accessibility (A11y) migration

chadsunvice:
Migrating the FEDS Accessibility checks.


chadsunvice:
@amauch-adobe  & @skumar09 please postpone review on this as I have some improvements and development ideas to make it reusable for other teams. You can test it our if you please though. It generates comprehensive A11y `.html` of violations found in our Milo pages.

I think Monday everything should be updated to the final version I am envisioning.

features/feds/a11y.spec.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  name: 'FEDS Accessibility',
+  features: [
+    {
+      name: '@A11y-on-FEDS-pages',
```
JackySun9, line 5: I think we should follow the way to write tests, check santosh's recent checkins

tests/feds/a11y.test.js
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable import/extensions */
+/* eslint-disable import/named */
+/* eslint-disable no-restricted-syntax */
+/* eslint-disable no-await-in-loop */
+import { FedsHeader } from '../../selectors/feds/feds.header.page';
+import a11y from '../../libs/accessibility';
+
+const AxeBuilder = require('@axe-core/playwright').default;
+const { expect, test } = require('@playwright/test');
+const parse = require('../../libs/parse.js');
```
JackySun9, line 10: We don't use parse again, use a new way

tests/feds/a11y.test.js
```diff
@@ -0,0 +1,48 @@
...
+      // Navigate to FEDS header page:
+      await page.goto(props.url);
+      // Wait for page to load & stabilize:
+      await page.waitForLoadState('networkidle');
+      await page.waitForTimeout(2500);
+      // Wait for FEDS GNAV to be visible:
+      await Header.MainNavContainer.waitFor({ state: 'visible', timeout: 5000 });
+
+      // Analyze page accessibility:
+      const a11yReport = await new AxeBuilder({ page })
```
JackySun9, line 29: Why don't put this into libs, it should be shared for others

JackySun9 (changes requested):

features/feds/a11y.spec.js
amauch-adobe: I agree with @JackySun9 here.

libs/accessibility.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable no-undef */
+/* eslint-disable no-unused-vars */
+/* eslint-disable max-len */
+/* eslint-disable import/no-import-module-exports */
+/* eslint-disable-next-line import/no-import-module-exports */
+const fs = require('fs').promises;
+const path = require('path');
+import { createHtmlReport } from 'axe-html-reporter';
+
+async function generateA11yReport(report, summary, project='feds', wcag='wcag21aa') {
```
amauch-adobe, line 10: Can we combine the a11y report with the current html-reporter so there is just one report?

package.json
```diff
@@ -25,7 +25,9 @@
   },
   "homepage": "https://github.com/adobecom/nala#readme",
   "dependencies": {
+    "@axe-core/playwright": "^4.7.0",
     "@playwright/test": "^1.32.2",
+    "axe-html-reporter": "^2.2.3",
```
amauch-adobe, line 30: If we combine the report with the current HTML report will we need the other dependency for the axe HTML reporter?

tests/feds/a11y.test.js
amauch-adobe: best to use the new standard than using parse.js.  @skumar09 or I need to update the wiki for the new method.

tests/feds/a11y.test.js
```diff
@@ -0,0 +1,48 @@
...
+        .withTags(['wcag2a', 'wcag2aa', 'wwcag21a', 'wcag21aa'])
+        .analyze();
+      // Assert page violations are limited:
+      expect(a11yReport.violations.length).toBeLessThan(5);
+
+      // Custom-summary for a11y html report:
+      const reportSummary =
+      `Report Summary: Checks the page accessibility of the given URL & generates a a11y report.
```
amauch-adobe, line 37: Might be best to standardize on an email that the whole team can subscribe to or is a part of just incase if team members change or you switch too. :) That way the communication doesn't get lost.

amauch-adobe: Without seeing what this report looks like, is there a need to have te custom report summary?  Just seems like a nice feature, but maybe not a necessary one.  Plus this could become difficult to maintain in adding to each and every test.

amauch-adobe (changes requested):
This is cool!  @skumar09, @JackySun9 and I had a brief meeting and discussed the need to have one report instead of another report that may be neglected.  We also felt that it would be good to have it associated with a '@accessibility' tag that could fire off these tests separately.  Lastly, we also considered that instead of having an entirely separate test created for accessibility it could instead be a tool that gets called on different blocks during their respective test(s), i.e. a test step that way the generic test report could receive the result.
Lastly, it could be fired off as a regression, daily, on-demand workflow execution within the CI/CD since someone could change the code outside of a PR.

Oh, and a last desire for accessibility would be a tool that could be run over current pages out in production to audit accessibility (although I do think there is a team that already has a tool and does this).

@chadsunvice, really good work, I like it.  Just a few recommendations and ideas.

amauch-adobe (changes requested):
This is cool!  @skumar09, @JackySun9 and I had a brief meeting and discussed the need to have one report instead of another report that may be neglected.  We also felt that it would be good to have it associated with a '@accessibility' tag that could fire off these tests separately.  Lastly, we also considered that instead of having an entirely separate test created for accessibility it could instead be a tool that gets called on different blocks during their respective test(s), i.e. a test step that way the generic test report could receive the result.
Lastly, it could be fired off as a regression, daily, on-demand workflow execution within the CI/CD since someone could change the code outside of a PR.

Oh, and a last desire for accessibility would be a tool that could be run over current pages out in production to audit accessibility (although I do think there is a team that already has a tool and does this).

@chadsunvice, really good work, I like it.  Just a few recommendations and ideas.

features/feds/a11y.spec.js
chadsunvice: I agree @JackySun9, @amauch-adobe. We will update these next week. We had some Footer/Jarvis component testing which took priority. This will be handled after those. Leave this open.

tests/feds/a11y.test.js
chadsunvice: Yes, working on this. See my previous comment.

chadsunvice:
> This is cool! @skumar09, @JackySun9 and I had a brief meeting and discussed the need to have one report instead of another report that may be neglected. We also felt that it would be good to have it associated with a '@accessibility' tag that could fire off these tests separately. Lastly, we also considered that instead of having an entirely separate test created for accessibility it could instead be a tool that gets called on different blocks during their respective test(s), i.e. a test step that way the generic test report could receive the result. Lastly, it could be fired off as a regression, daily, on-demand workflow execution within the CI/CD since someone could change the code outside of a PR.
> 
> Oh, and a last desire for accessibility would be a tool that could be run over current pages out in production to audit accessibility (although I do think there is a team that already has a tool and does this).
> 
> @chadsunvice, really good work, I like it. Just a few recommendations and ideas.

Thanks @amauch-adobe! I reopened the PR and this and the next week I'm going to focus on incorporating your suggestions. We have a KPI to deliver A11y for our components and I love the idea of having this as a daily regression that would trigger warnings if there are multiple WCAG violations found. We could sync on this tomorrow at the NALA bi-weekly.

skumar09:
> > This is cool! @skumar09, @JackySun9 and I had a brief meeting and discussed the need to have one report instead of another report that may be neglected. We also felt that it would be good to have it associated with a '@accessibility' tag that could fire off these tests separately. Lastly, we also considered that instead of having an entirely separate test created for accessibility it could instead be a tool that gets called on different blocks during their respective test(s), i.e. a test step that way the generic test report could receive the result. Lastly, it could be fired off as a regression, daily, on-demand workflow execution within the CI/CD since someone could change the code outside of a PR.
> > Oh, and a last desire for accessibility would be a tool that could be run over current pages out in production to audit accessibility (although I do think there is a team that already has a tool and does this).
> > @chadsunvice, really good work, I like it. Just a few recommendations and ideas.
> 
> Thanks @amauch-adobe! I reopened the PR and this and the next week I'm going to focus on incorporating your suggestions. We have a KPI to deliver A11y for our components and I love the idea of having this as a daily regression that would trigger warnings if there are multiple WCAG violations found. We could sync on this tomorrow at the NALA bi-weekly.

@chadsunvice @amauch-adobe @JackySun9 

In the past few days, I have been exploring on this offline, I have a few additional thoughts along this, will schedule a call and 
let us discuss them. ( we can use our regular call that is tomorrow for this)

tests/feds/a11y.test.js
chadsunvice: Updated!

chadsunvice:
PR updated to latest requirements. I've added a few `test.step` accessibility examples in the `/feds/` path. Every client will have to adjust the step to match their critical components. I would argue, it is only required for big, important blocks and shouldn't be abused and copy-pasted into every test footer as a closing step. 

eg: HEADER/FOOTER make sense for us to be always checks as blocks, but Search, doesn't as it is, more or less, already analysed in the HEADER a11y checks.

At the end of the test, the JSON report is attached to each test results instance, giving the NALA uses the chance to debug per block/component.  

Screenshot 2023-07-27 at 16 14 27

Screenshot 2023-07-27 at 16 15 05

libs/accessibility.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable max-len */
+/* eslint-disable no-undef */
+/* eslint-disable no-unused-vars */
+/* eslint-disable import/no-import-module-exports */
+/* eslint-disable import/no-extraneous-dependencies */
+import * as path from 'path';
```
JackySun9, line 6: In the context of Node.js's path module, you don't need to use the * because the path module exports a single object with methods for working with file paths. This means that you can import the path module as a whole using either ES6 or CommonJS syntax: 

ES6 (for Node.js >=14):
import path from 'node:path';

libs/accessibility.js
```diff
@@ -0,0 +1,53 @@
...
+import { promises as fs } from 'fs';
+import { createHtmlReport } from 'axe-html-reporter';
+
+async function generateA11yReport(report, summary, project = 'feds', wcag = 'wcag21aa') {
+  // Timestamp reports to track & avoid duplicates:
+  const time = new Date();
+  const crtEnv = 'prod';
```
JackySun9, line 13: You hardcode env here, what does mean for this one?

package.json
```diff
@@ -25,13 +25,15 @@
...
-    "@playwright/test": "^1.32.2",
+    "@axe-core/playwright": "^4.7.0",
+    "axe-html-reporter": "^2.2.3",
     "axios": "^1.3.4",
     "js-yaml": "^4.1.0"
   },
   "devDependencies": {
     "@babel/core": "7.21.3",
     "@babel/eslint-parser": "7.21.3",
+    "@playwright/test": "^1.36.2",
```
JackySun9, line 36: You updated playwright package, this might impact all other tests

tests/feds/a11y.test.js
```diff
@@ -0,0 +1,54 @@
...
+        await page.waitForTimeout(5000);
+        // Check the expected URL was loaded:
+        await expect(page).toHaveURL(`${baseURL}${a11yUrl}${features[0].browserParams}`);
+      });
+
+      await test.step('Analyze targeted page accessibility', async () => {
+        // Wait for FEDS GNAV to be visible:
+        await Header.mainNavContainer.waitFor({ state: 'visible', timeout: 5000 });
+        // Analyze page accessibility:
+        a11yReport = await new AxeBuilder({ page })
```
JackySun9, line 33: Maybe move this to accessibility.js, then if any change in future doesn't impact *.test.js files, also provide append function to keep collect violations when user does action on the pages

tests/feds/a11y.test.js
```diff
@@ -0,0 +1,54 @@
...
+        // Custom-summary for a11y html report:
+        const reportSummary = `Report Summary: Checks the page accessibility 
+            of the given URL & generates a a11y report.
+            <br>More info:</br>
+            <ol style="margin: 0">
+            <li>This is a Federated Services, A11y initiative.</li>
+            </ol>
+          `;
+        // Generate a11y html report:
+        await a11y.generateA11yReport(a11yReport, reportSummary, 'feds');
```
JackySun9, line 50: by the way, how to see the report, local?

tests/feds/consent.test.js
```diff
@@ -67,5 +70,20 @@ test.describe('Consent Component test suite', () => {
       // Check FEDS browser objects (post-consent):
       await Consent.assertOneTrustCookieGroups(1);
     });
+
+    await test.step('Analyze consent block accessibility', async () => {
+      // Analyze page accessibility:
+      const a11yReport = await new AxeBuilder({ page })
```
JackySun9, line 76: This is really cool, maybe move to accessibility.js

tests/feds/consent.test.js
```diff
@@ -67,5 +70,20 @@ test.describe('Consent Component test suite', () => {
...
+        .withTags(['wcag2a', 'wcag2aa', 'wwcag21a', 'wcag21aa'])
+        .include('#onetrust-banner-sdk')
+        .analyze();
+      // Assert there are no page accessibility violations:
+      expect.soft(a11yReport.violations.length).toBeLessThan(5);
+      // Attach A11y results to test output:
+      await testInfo.attach('a11y-scan-results', {
```
JackySun9, line 83: Why do we need this if we have html report

JackySun9 (changes requested):

tests/feds/a11y.test.js
chadsunvice: @JackySun9, the clients can import the A11Y library at their will and use it. I think it is overkill to pass the `page` instance to an `accessibiliity` method, that returns a report for you, when this it just a single line of code.

libs/accessibility.js
chadsunvice: Indeed, I will remove this, good find. It was a placeholder for having dynamic report naming based on test environments.

tests/feds/consent.test.js
chadsunvice: Yes, this attaches the `json` ONLY, to the reporters, so clients can take a look when debugging for other things, if there are any a11y violations.

tests/feds/consent.test.js
chadsunvice: I'll wait to see what @skumar09 & @amauch-adobe have to say too and we'll decide. But I don't think it should be a method specifically. But, we can accommodate anything at this point.

tests/feds/a11y.test.js
chadsunvice: Go to `/axe-reports` in your local project, reveal in Finder, open HTML reports with Google Chrome, or your browser of choice.

package.json
chadsunvice: I can remove this, yes, this has no effect on the logic, or the checks. It was just a test.

tests/feds/a11y.test.js
```diff
@@ -0,0 +1,54 @@
...
+        // Check the expected URL was loaded:
+        await expect(page).toHaveURL(`${baseURL}${a11yUrl}${features[0].browserParams}`);
+      });
+
+      await test.step('Analyze targeted page accessibility', async () => {
+        // Wait for FEDS GNAV to be visible:
+        await Header.mainNavContainer.waitFor({ state: 'visible', timeout: 5000 });
+        // Analyze page accessibility:
+        a11yReport = await new AxeBuilder({ page })
+          .withTags(['wcag2a', 'wcag2aa', 'wwcag21a', 'wcag21aa'])
```
skumar09, line 34: @chadsunvice: 
 - I am not 100% sure of using all the accessibility tags for our applications. ( `wcag21aa` is sufficient I belive)
 - Also, I think these tags can go to `.spec.js/ test cases` WDYT?

tests/feds/consent.test.js
```diff
@@ -67,5 +70,20 @@ test.describe('Consent Component test suite', () => {
       // Check FEDS browser objects (post-consent):
       await Consent.assertOneTrustCookieGroups(1);
     });
+
+    await test.step('Analyze consent block accessibility', async () => {
+      // Analyze page accessibility:
+      const a11yReport = await new AxeBuilder({ page })
+        .withTags(['wcag2a', 'wcag2aa', 'wwcag21a', 'wcag21aa'])
+        .include('#onetrust-banner-sdk')
```
skumar09, line 78: - having accessibility tests for particular sections is best suited for blocks, 
- `'#onetrust-banner-sdk'`selector/locator can come from pom, as all our locators are in pom

tests/feds/consent.test.js
skumar09, line 83: - I think `testInfo.attach('a11y-scan-results')` along with soft validation`expect.soft(a11yReport.violations.length).toBeLessThan(5)`; is a good and handy option, till our accessibility tests report becomes a mandatory

tests/feds/a11y.test.js
```diff
@@ -0,0 +1,54 @@
...
+          .analyze();
+        // Assert page violations are limited:
+        expect(a11yReport.violations.length).toBeLessThan(5);
```
skumar09, line 37: - you may want to have this as a `soft expect`

skumar09 (changes requested):
@chadsunvice: 
accessibility PR is looking, please find below few review points / thoughts, 
- I think as of now, we may not need all accessibility testing tags [ `'wcag2a', 'wcag2aa', 'wwcag21a', 'wcag21aa'`], I think 'wcag21aa' is good enough, and I think this tag can go to spec or any global variable
- we can have accessibility testing for a page or block/section, depending on the corresponding project/spec(test case)
- I like attaching accessibility test summary as testInfo, and I think soft expect() and testInfo attachment will meet the current need.
- Let us not upgrade the playwright version ( @JackySun9 is a comment)
- few minor comments inline in the page/code line

cc : @JackySun9  @amauch-adobe

tests/feds/a11y.test.js
amauch-adobe: I agree with @chadsunvice on this, if it is just an import call and class builder call then there is no reason to create a wrapper class unless of course, we start adding our own methods, etc. that would require better util grouping.

libs/accessibility.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable max-len */
+/* eslint-disable no-undef */
+/* eslint-disable no-unused-vars */
+/* eslint-disable import/no-import-module-exports */
+/* eslint-disable import/no-extraneous-dependencies */
+import * as path from 'path';
+import { promises as fs } from 'fs';
+import { createHtmlReport } from 'axe-html-reporter';
+
+async function generateA11yReport(report, summary, project = 'feds', wcag = 'wcag21aa') {
```
amauch-adobe, line 10: I know Santosh mentioned only using one tag type for "wcag"..., one way or another, if we do go with just one type would the "wcag" parameter need to coincide with the tag as well?

package-lock.json
amauch-adobe, line 1: make sure you rebuild the package-lock.json after you downgrade the playwright version as requested by Jacky.

tests/feds/consent.test.js
amauch-adobe: I don't know if there would be a reason to move a one-liner call to a util class to just use a wrapper method one-liner to replace it.

amauch-adobe (commented):
See my comments inline, I'm good for approval once changes from others are addressed.

JackySun9:
@chadsunvice did you update the code?

amauch-adobe:
> @chadsunvice did you update the code? I also didn't see any changes needing review from my last review.  Was there supposed to be?

amauch-adobe (commented):
I didn't see any changes since your last commits.

libs/accessibility.js
chadsunvice: Done!

tests/feds/a11y.test.js
chadsunvice: As agreed with @amauch-adobe, will defer this ask as any client can import the lib and use it as they chose. It's just a single line of code.

tests/feds/a11y.test.js
chadsunvice: I really like the idea of having the `wcag` tags imported from the `spec` files @skumar09! As for the right tags to use, I have to do more research, and I'd also try to invite an Adobe A11y Champion in our next bi-monthly meeting to make sure we're taking the best decision on this. Will keep the config as is, but import it from `spec`s for now.

tests/feds/consent.test.js
chadsunvice: Fixed/updated!

libs/accessibility.js
```diff
@@ -0,0 +1,53 @@
...
+  // Write a11y html report to file:
+  await fs.writeFile(path.resolve(folderPath, reportName), htmlReport, function(err) {
+    if (err) {
+      return console.error(`[FEDsError] Failed to write the '${reportName}' report! | Reason: '${err}'`);
+    }
+    console.info(`[FEDsInfo] Successfully wrote the '${reportName}' a11y report!`);
+  });
+}
+
+module.exports = { generateA11yReport };
```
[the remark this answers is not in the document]
chadsunvice: I tackled most of these, if not all of them. Marking this as resolved.

chadsunvice:
I updated the PR and addressed most of, if not all the concerns. I've also fixed a few issues that I found while testing the feature. See updates.

Let's get this merged and we can tackle improvements in other PRs. This PR conversation is hard to follow with all the reviews and messages.

Cc: @amauch-adobe, @skumar09, @JackySun9.

JackySun9 (approved):

amauch-adobe (approved):
Approved.  Sorry for taking so long and holding you up.  Excellent job!

PR #123, opened 2023-05-19
Add analytics test for uar

JackySun9:
https://jira.corp.adobe.com/browse/MWPW-126153
1. add utils for analytics test
2. add uar as sample for analytics test
3. this is only working for firefox


tests/uar/quiz.analytics.test.js
```diff
@@ -0,0 +1,64 @@
...
+      },
+    );
+  }
+
+  test.afterAll(async () => {
+    console.log('After all tests: Disable network logging');
+    // Disable network logging
+    webUtil.disableNetworkLogging();
+  });
+});
```
skumar09, line 64: @JackySun9 : trying to understand the context for `capturing network logs`, are we using it for some validations?

tests/uar/quiz.analytics.test.js
```diff
@@ -0,0 +1,64 @@
...
+    networklogs = [];
+    page = await browser.newPage();
+    webUtil = new WebUtil(page);
+    console.log('Before all tests: Enable network logging');
+    // Enable network logging
+    webUtil.enableNetworkLogging(networklogs);
+  });
+
+  // reset timeout because we use this to run all test data
+  test.setTimeout(3 * 60 * 1000);
```
skumar09, line 27: - you may want to move `test.setTimeout()` inside test() function/method
- if you need 3 times of default timeout, you may want to use `test.slow()` as below or you can use `test.beforeEach()` to setTimeout
```
- test('slow test', async ({ page }) => {
  test.slow(); 
  // ...
});
```

libs/webutil.js
```diff
@@ -222,4 +222,28 @@ exports.WebUtil = class WebUtil {
...
+  /**
+   * Enable network logging
+   * @param {Array} networklogs - An array to store all network logs
+   */
+  async enableNetworkLogging(networklogs) {
+    await this.page.route('**', (route) => {
+      const url = route.request().url();
+      if (url.includes('sstats.adobe.com/ee/or2/v1/interact') || url.includes('sstats.adobe.com/ee/or2/v1/collect')) {
+        networklogs.push(url);
+        // eslint-disable-next-line max-len, no-underscore-dangle
```
skumar09, line 235: - You want to split the code line, to avoid adding // eslint-disable-next-line max-len
- I know we have this on many other tests, and i am trying to format them..

features/uar/quiz.analytics.spec.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  name: 'Quiz Recommender Blocks',
+  features: [
+    {
+      tcid: '0',
+      name: '@quiz single template',
+      path: '/creativecloud/quiz-recommender.html',
+      tags: '@quiz-analytics-single-template @quiz-analytics',
```
skumar09, line 8: - In tags you may want to include the project name, something like `cc or uar` kind.

tests/uar/quiz.analytics.test.js
JackySun9: Currently I just catch the real logs and put them into an array, I am thinking where to put the expected data and add the validation part

libs/webutil.js
JackySun9: how to do this?

features/uar/quiz.analytics.spec.js
JackySun9: sure, will do

amauch-adobe (changes requested):
Approved, just address @skumar09's comments

features/uar/quiz.spec.js
```diff
@@ -5,63 +5,63 @@ module.exports = {
       tcid: '0',
       name: '@quiz single template',
       path: '/creativecloud/quiz-recommender.html',
-      tags: '@quiz-single-template @quiz-static',
+      tags: '@uar-quiz-single-template @uar-quiz-static',
```
chadsunvice, line 8: I suggest we use some team-specific tags for each test as to have a way to filter checks/run regressions from team-perspective too. (eg: `tags: '@cc @uar-quiz-single-template @uar-quiz-static',`)

features/uar/quiz.spec.js
```diff
@@ -5,63 +5,63 @@ module.exports = {
...
       data: 'data/uar/quiz/quiz-single-template.yml',
     },
     {
       tcid: '1',
       name: '@quiz double template',
       path: '/creativecloud/quiz-recommender.html',
-      tags: '@quiz-double-template @quiz-static',
+      tags: '@uar-quiz-double-template @uar-quiz-static',
```
chadsunvice, line 15: I suggest we use some team-specific tags for each test as to have a way to filter checks/run regressions from team-perspective too. (eg: `tags: '@cc @uar-quiz-single-template @uar-quiz-static',`)

Same goes for all other test tag lines.

chadsunvice (approved):
Looks good Jacky! 👍

libs/webutil.js
```diff
@@ -222,4 +222,30 @@ exports.WebUtil = class WebUtil {
...
+      route.continue();
+    });
+  }
+
+  /**
+   * Disable network logging
+   */
+  async disableNetworkLogging() {
+    await this.page.unroute('**');
+    await this.page.close();
```
skumar09, line 249: - do we need `page.close()` to disable network logging
- we may have a scenario where we just want to capture network logs for a single navigation/click or single test step.
- so calling this method would close the page.

skumar09 (changes requested):

tests/uar/quiz.analytics.test.js
```diff
@@ -0,0 +1,65 @@
...
+          });
+
+          // test step-2
+          // eslint-disable-next-line no-await-in-loop
+          await test.step(`Check results on test page according to ${testdata[key]}`, async () => {
+            await quiz.checkResultPage(testdata[key]);
+          });
+        }
+
+        console.log(networklogs);
```
skumar09, line 55: - this would a lot of logging to the terminal, I would suggest commenting or removing this statement if there is no validation of logs done at the moment

libs/webutil.js
JackySun9: yes, I removed the close() method

tests/uar/quiz.analytics.test.js
JackySun9: ok

skumar09 (approved):

amauch-adobe (approved):
Approved.

PR #127, opened 2023-06-03
github CI/CD for visual testing

amauch-adobe:
Addresses: [MWPW-132172](https://jira.corp.adobe.com/browse/MWPW-132172)
- added github actions for visual testing executions in the CI/CD pipeline


mokimo:
Is visual testing giving feedback / screenshots on what didn't work ? Or what does that do

amauch-adobe:
@mokimo > Is visual testing giving feedback / screenshots on what didn't work ? Or what does that do

Visual testing checks by pixel ratio (based on the configuration for how strict) a page's content or a block of content has changed, if any.  The method HasScreenshots first checks if there are any screenshots already recorded for comparison checking.  If not it creates some (basically if a new test has been created), if there are screenshots found it will then take a subset of screenshots for comparison checking.  Based on the pixel ratio strictness configured if there has been a change of content to the page or element on the page being checked the test will fail, basically stating that the screenshots are not the same anymore.  Visual testing allows you to check on different platforms (OS), browsers, and devices to make sure brand content hasn't changed.  Say if an author makes an accidental change or content goes up missing or isn't rendering when this test runs it will find it is missing, has changed, etc.  If a change was intended then we as executors would just need to update the saved screenshots to the new baseline for the page/element that was changed for subsequent runs.

Now obviously small tiny pixel shifts happen in the rendering of an element or page which doesn't necessarily mean the test is a failure since the item or page's content could still be just fine just slightly shifted.  This is why the configuration will need to be adjusted to find a reasonable strictness level.  See [here for reference.](https://playwright.dev/docs/test-snapshots)

amauch-adobe:
closing this until I or someone is ready to tackle it again.

PR #129, opened 2023-06-07
MWPW-132286 : Accordion block tests scripts

skumar09:
[MWPW-132286](https://jira.corp.adobe.com/browse/MWPW-132286): Accordion block tests scripts
PR includes:
- Automation test scripts for the following features of the Accordion block
  - Accordion
  - Accordion (SEO)
  - Accordion (Quiet, max-width-12-desktop-large)


selectors/milo/accordion.block.page.js
```diff
@@ -0,0 +1,178 @@
...
+
+    // verify accordion headers, buttons, and icons count
+    await expect(await this.accordionHeaders).toHaveCount(data.headers);
+    await expect(await this.accordionButtons).toHaveCount(data.headers);
+    await expect(await this.accordionButtonIcons).toHaveCount(data.headers);
+
+    // verify accordion headers text content
+    await expect(await this.accordionHeaders.nth(0)).toContainText(data.heading0);
+    await expect(await this.accordionHeaders.nth(1)).toContainText(data.heading1);
+    await expect(await this.accordionHeaders.nth(2)).toContainText(data.heading2);
```
chadsunvice, line 106: I think the awaits inside the expect statement are redundant and have no effect. Other selector pages using same style should also be changed.

**eg:**

```
    await expect(this.accordionHeaders).toHaveCount(data.headers);
    await expect(this.accordionButtons).toHaveCount(data.headers);
    await expect(this.accordionButtonIcons).toHaveCount(data.headers);
```

features/milo/accordion.block.spec.js
```diff
@@ -0,0 +1,43 @@
+module.exports = {
+  FeatureName: 'Accordion Block',
+  features: [
+    {
+      tcid: '0',
+      name: '@Accordion ',
```
chadsunvice, line 6: Training whitespace in `name` feature attribute.

features/milo/accordion.block.spec.js
```diff
@@ -0,0 +1,43 @@
...
+        heading0: 'How do I compress a PDF without losing quality?',
+        heading1: 'What size PDFs can I compress?',
+        heading2: 'How do I check my PDF file size?',
+      },
+      tags: '@accordion @smoke @regression @milo,',
+    },
+
+    {
+      tcid: '1',
+      name: '@Accordion (seo) ',
```
chadsunvice, line 19: Training whitespace in `name` feature attribute.

features/milo/accordion.block.spec.js
```diff
@@ -0,0 +1,43 @@
...
+        heading0: 'How do I compress a PDF without losing quality?',
+        heading1: 'What size PDFs can I compress?',
+        heading2: 'How do I check my PDF file size?',
+      },
+      tags: '@accordion @accordion-seo @smoke @regression @milo,',
+    },
+
+    {
+      tcid: '2',
+      name: '@Accordion (quiet, max-width-12-desktop-large) ',
```
chadsunvice, line 32: Training whitespace in `name` feature attribute.

selectors/milo/accordion.block.page.js
```diff
@@ -0,0 +1,178 @@
...
+export default class Accordion {
+  constructor(page) {
+    this.page = page;
+
+    // accordion  locators
+    this.accordion = this.page.locator('.accordion-container');
+    this.accordionForeground = this.accordion.locator('.foreground');
+    this.accordionHeaders = this.accordion.locator('dt[role=heading]');
+    this.accordionButtons = this.accordion.locator('dt button');
+    this.accordionButtonIcons = this.accordion.locator('.accordion-icon');
```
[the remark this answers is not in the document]
skumar09: - It is a playwright feature:  `chain locators`, and it is a best practice.
- In chain locators, you can locate a parent element and then search for sub-elements within its scope, which is an efficient approach to target specific elements within a web page,  by narrowing down the search to a specific element, you can reduce the search scope.
- Here, in the above example,  it is not required to search all locators within the entire page scope, ( `page.locator()` ), as we know they are sub-elements of the accordion block, so once you locate the accordion block in the page, then the sub-elements of the accordion can be searched with its reference. 
- on changing element selectors, 
  - I think, if the accordion changes then we need to update only one place of change, which is a good thing 
-  It is a kind of new feature w.r.t other automation tools, and I think we have to help people to know about it.

selectors/milo/accordion.block.page.js
skumar09: - this is a bit interesting, we are using these await inside `async methods (){}` as required
- I think, In the Playwright the `expect()` function already handles promises, so you can directly pass the promise returned by (in this case `this.accordionHeaders`) to `expect()`. And the `expect()` function will automatically wait for the promise to resolve before performing the assertion. so await outside the `expect()` (not inside) becomes redundant. But I have seen a few assertions failing when I add await outside the `expect()` and not inside.
- In this case, I have also I tried removing await,  and a combination of  'only inside' or ' only outside', in all cases the tests are passing. And for the `.toBeTruthy()`, assertion, sometime assertions are failing with outside await.
- I think I need to do further due diligence on this, I have created an [Jira](https://jira.corp.adobe.com/browse/MWPW-132348) for the same,

chadsunvice: I agree with your reasoning @skumar09. From my testing, I've seen the need to use `await` when the expected operator is a function which returns a promise. Thus, we either use the inside variant, or the outside variant. Having a static expect value, doesn't require waiting.

My mention was due to consistency with the rest of the selector pages, and matching the documentation examples: https://playwright.dev/docs/api/class-locatorassertions.

chadsunvice (approved):
Looks good! ➕

JackySun9 (approved):

Dli3 (approved):

PR #137, opened 2023-06-26
[MWPW-131179] Added Jarvis NALA checks & supporting logic

iamdanchiv:
PR relates to story [MWPW-131179](https://jira.corp.adobe.com/browse/MWPW-131179): Develop FEDS Jarvis checks on NALA.


iamdanchiv:
Closing to activate Adobe Signed CLA. Re-opening after.

tests/feds/jarvis.desktop.test.js
```diff
@@ -0,0 +1,99 @@
+/* eslint-disable import/named */
+/* eslint-disable no-await-in-loop */
+/* eslint-disable import/extensions */
+import { expect, test } from '@playwright/test';
+import { FedsJarvis } from '../../selectors/feds/feds.jarvis.page.js';
+import { FedsConsent } from '../../selectors/feds/feds.consent.page.js';
+import * as JarvisSpec from '../../features/feds/jarvis.spec.js';
```
JackySun9, line 7: We don't want star here

JackySun9 (changes requested):

tests/feds/jarvis.desktop.test.js
amauch-adobe: Same concern mentioned in my comments on your PR #138.  Sorry again for not checking that out before.

tests/feds/jarvis.mobile.test.js
```diff
@@ -0,0 +1,49 @@
+/* eslint-disable import/named */
+/* eslint-disable no-await-in-loop */
+/* eslint-disable import/extensions */
+import { expect, test } from '@playwright/test';
+import { FedsJarvis } from '../../selectors/feds/feds.jarvis.page.js';
+import { FedsConsent } from '../../selectors/feds/feds.consent.page.js';
+import * as JarvisSpec from '../../features/feds/jarvis.spec.js';
```
amauch-adobe, line 7: The same here on the need to use require() instead of "import *" where we are using CommonJS modules.export structure for spec files.  I think it would be provincial for us to look into seeing if updating spec files as separate export declarations would be a viable option so we could be more ES6 compliant, though.

amauch-adobe (changes requested):
Everything looks great, just that one concern of using "import *" with our spec.js files since it fails runtime execution as mentioned in PR #138

tests/feds/jarvis.desktop.test.js
iamdanchiv: I've read your concerns @amauch-adobe from [PR #138](https://github.com/adobecom/nala/pull/138#pullrequestreview-1501665646), and I don't fully agree. I think we should start the migration to ES6 right away.

I updated the scripts to import `{ features }` instead of `*`, and I hope this is an agreeable middle ground. I've also ran the checks multiple times in several envs and they never fail. Can you provide runs where this approach fails? I was unable to reproduce a scenario where these fail, especially since we use imports already.

Please give palpable examples as to why we shouldn't use `imports` everywhere and still use CommonJS logic. Also the fact that this is 100% asynchronous should support the ethos of a PlayWright framework even more.

Cc: @narcis-radu, @skumar09, @JackySun9

tests/feds/jarvis.mobile.test.js
iamdanchiv: I've read your concerns @amauch-adobe from [PR #138](https://github.com/adobecom/nala/pull/138#pullrequestreview-1501665646), and I don't fully agree. I think we should start the migration to ES6 right away.

I updated the scripts to import `{ features }` instead of `*`, and I hope this is an agreeable middle ground. I've also ran the checks multiple times in several envs and they never fail. Can you provide runs where this approach fails? I was unable to reproduce a scenario where these fail, especially since we use imports already.

Please give palpable examples as to why we shouldn't use `imports` everywhere and still use CommonJS logic. Also the fact that this is 100% asynchronous should support the ethos of a PlayWright framework even more.

Cc: @narcis-radu, @skumar09, @JackySun9

iamdanchiv:
I've read your concerns @amauch-adobe from [PR #138](https://github.com/adobecom/nala/pull/138#pullrequestreview-1501665646), and I don't fully agree. I think we should start the migration to ES6 right away.

I updated the scripts to import `{ features }` instead of `*`, and I hope this is an agreeable middle ground. I've also ran the checks multiple times in several envs and they never fail. Can you provide runs where this approach fails? I was unable to reproduce a scenario where these fail, especially since we use imports already.

Please give palpable examples as to why we shouldn't use `imports` everywhere and still use CommonJS logic. Also the fact that this is 100% asynchronous should support the ethos of a PlayWright framework even more.

Cc: @narcis-radu, @skumar09, @JackySun9

selectors/feds/feds.jarvis.page.js
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable import/no-import-module-exports */
+
+exports.FedsJarvis = class FedsJarvis {
```
narcis-radu, line 3: should use `module.exports`

narcis-radu (approved):

PR #138, opened 2023-06-27
[MWPW-131179] Remove parse.js logic from FEDS checks

iamdanchiv:
PR related to [MWPW-132768](https://jira.corp.adobe.com/browse/MWPW-132768): Remove parse.js logic from FEDS checks

**Acceptance Criteria:**

* update FEDS checks logic to not use parse.js library
* create additional logic supporting the FEDS checks where required
* test files affected:
  * userprofile.test.js
  * consent.test.js
  * header.test.js
  * search.test.js
* linting/refactoring some Milo checks


amauch-adobe (changes requested):
@iamdanchiv, Santosh and I had a quick conversation about this just before last weekend and Santosh found out that the "import *..." command for the spec.js files causes tests to fail.  Is that something that is happening on your side of local testing?    

I apologize for quickly not checking this before when I approved the PR in your prior closed one.  I figured you were moving to the import/export method of doing things to be compliant with ES6 over CommonJS which uses require/module.exports style.  In a general sense we do want to follow ES6 since I believe Milo's code base is standardizing on it.  We might need to first create a ticket to update how specs are done so that they could work with the import nomenclature since "import" calls are asynchronous and provide potential performance improvements over "require".  

Santosh and I think where the issue might be is how the spec files use module.export calls rather than have separate export method declarations like how our POM files are.  This is why we think the "import *..." call won't work but still needs to remain a require() call instead.

That being said everything else looks good to me and is really great work!

tests/milo/accordion.block.test.js
```diff
@@ -1,73 +1,66 @@
...
+test.describe('Milo Accordion Block test suite', () => {
   test.beforeEach(async ({ page }) => {
     accordion = new Accordion(page);
   });
 
   // Test - 1
-  test(`${features[0].name}, @milo-live, ${features[0].tags},https://milo.adobe.com `, async ({ page, baseURL }) => {
-    console.info('Base URL test :  ', baseURL);
+  test(`${features[0].name} › @milo-live › ${features[0].tags} › https://milo.adobe.com`, async ({ page, baseURL }) => {
+    console.info('[MiloInfo] Checking page: ', baseURL);
```
skumar09, line 17: - Please revert back this change to the comma separated.
- context: It has an impact on the report parser, ( as we use comma separation logic there)

tests/milo/accordion.block.test.js
```diff
@@ -1,73 +1,66 @@
...
     await test.step('Go to Accordion block test page', async () => {
       await page.goto(`${baseURL}${features[1].path}`);
       await page.waitForLoadState('domcontentloaded');
-
       await expect(page).toHaveURL(`${baseURL}${features[1].path}`);
     });
 
     // test step-2
-    await test.step('Verify Accrodion block content / specs ', async () => {
+    await test.step('Verify Accrodion block content/specs', async () => {
```
skumar09, line 45: This is my view, and request others to take a look and provide views.
- The verbiage related to the test case or step name, description,  capital or non-capital, space or no space between separators, etc. I would leave it to individual projects 
- we can correct for any spell check or any syntax errors or format errors
cc : @amauch-adobe  @JackySun9

tests/milo/carousel.block.test.js
```diff
@@ -1,18 +1,16 @@
-/* eslint-disable import/extensions */
 /* eslint-disable import/named */
+/* eslint-disable import/extensions */
 import { expect, test } from '@playwright/test';
 import { Carousel } from '../../selectors/milo/carousel.block.page';
-
-const CarouselSpec = require('../../features/milo/carousel.block.spec');
+import * as CarouselSpec from '../../features/milo/carousel.block.spec';
```
skumar09, line 5: - I don't think we need `import * `, as .spec.js exports only a single module object.

cc : @amauch-adobe  @JackySun9

skumar09 (changes requested):

tests/milo/accordion.block.test.js
amauch-adobe: @iamdanchiv, I forgot about this in my review, @skumar09 is right the reporting parsers found under /utils/reporters require a certain amount of comma-separate titles since the Nala Dashboard Application created by Jingle expects it in the JSON payload.

tests/milo/accordion.block.test.js
amauch-adobe: I agree that test step naming, descriptions, etc. are up to the individuals. I don't think we need to set a specific standard since we aren't using a framework that requires unique/shared steps.  That being said I do believe grammar, syntax, and format errors should be reviewed during PR reviews.  Also, if the test step name/description doesn't make sense or provide enough clarity to a person for what is happening in the step method, that should also be called out with a provided optional change to improve it.

tests/milo/carousel.block.test.js
amauch-adobe: Yeah, we shouldn't use the import * method for the spec files due to the usage of module.exports = {}; in these files.  require(); is necessary due to it being CommonJS rather than ES6 standards.  Although as mentioned above if Milo is standardizing on ES standards then we should as well IMO.  Maybe we should create a ticket to look into a way to convert specs.js files to ES standards.

chadsunvice:
Re-opened the PR from the main account. Closing this one!

PR #141, opened 2023-07-12
MWPW-133693: Milo test scripts : update to ES6 way, refactor, code cleanup, comment cleanup, file name update

skumar09:
[MWPW-133693 ](https://jira.corp.adobe.com/browse/MWPW-133693): Milo test scripts update to ES6 way, refactor, code cleanup, comment cleanup, file name update
PR includes

- fix or remove the following `eslint` lines from test files
  - `/* eslint-disable max-len */`
  - `/* eslint-disable quote-props */`
  - `/* eslint-disable import/named */`
  - `/* eslint-disable import/extensions */`
  - `/* eslint-disable import/no-import-module-exports */`
- Update comments and cleanup redundant comments
   - As all contributors understand the Nala test.js template, we don't need basic comments anymore, 
   - cleaning/updating multiline comments.
- introducing import ordering
   - `playwright libs`
   - `Nala libs`
   - `spec.js`
   - `page.js `
-  `export default class` to  `export class` i.e. named export
    - fixing the discrepancies in the pom class export syntax.
    - we will use named export 
- Add step identifier in `test.step()` 
    - this is not mandatory for all projects, it is for individual project owners 
- Update `obj` variable name to `modal` a meaningful name name in geo routing tests ( pending review comment) 
- Moved `failed block spec/selectors/tests` to bacom folder, as they are running on `bacom pages`
- updated filename , tags and variable naming for htmlextn ( just to follow uniformity)

Note: 
- working on following block/features tests to move regular template from earlier parse logic
  - faas
  - htmlextn
  - imslogin


tests/bacom/failed.block.test.js
amauch-adobe, line 1: same concern here, I feel this test should stay under milo or at least a different directory for global tests or it should be separated for each.  Though the reason for keeping it under milo when it was first created was to test if a milo code change caused any consumer's pages to have failed blocks.

amauch-adobe (changes requested):
Looks like some of the tests are failing, we might need to add a specific test timeout for the htmlExtn test.  Also, the failed.block (test, spec, selectors) file I believe shouldn't be moved to the bacom folder since there are more than just the bacom urls being tested, i.e., (stock, DC).

tests/bacom/failed.block.test.js
skumar09: @amauch-adobe: It is interesting, I agree with your point, I think we should have failed block test scripts for each application(s) as per their pages.

amauch-adobe (approved):
Approved, based on @skumar09 and I's conversation in Nala Refinement about having the Failed Block test be a part of separate application teams' directories for ownership.

chadsunvice (approved):
LGTM!

tests/bacom/failed.block.test.js
```diff
@@ -1,26 +1,23 @@
...
 /* eslint-disable no-await-in-loop */
-import { WebUtil } from '../../libs/webutil';
+import { expect, test } from '@playwright/test';
+import { WebUtil } from '../../libs/webutil.js';
 
-const { expect, test } = require('@playwright/test');
 const parse = require('../../libs/parse.js');
-const failedBlock = require('../../features/milo/failedblock.spec.js');
-const selectors = require('../../selectors/milo/failedblock.selectors.js');
+const failedBlock = require('../../features/bacom/failed.block.spec.js');
```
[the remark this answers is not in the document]
skumar09: - These one  use require(), because they are using the different implementations, these will be changed once we move away from parse.js,

JackySun9 (approved):

PR #142, opened 2023-07-14
[MWPW-132769] Added breadcrumbs checks & supporting logic

chadsunvice:
PR relates to ticket [[NALA-Automation] Automate FEDS Breadcrumbs component on NALA](https://jira.corp.adobe.com/browse/MWPW-132769).


tests/feds/breadcrumbs.test.js
```diff
@@ -0,0 +1,104 @@
...
+test.describe('Breadcrumbs Component test suite', () => {
+  // FEDS Breadcrumbs-With-Base Checks:
+  test(`${features[0].name}, ${features[0].tags}`, async ({ page, baseURL }) => {
+    const Header = new FedsHeader(page);
+    console.info(`[FEDSInfo] Checking page: ${baseURL}${features[0].path}`);
+
+    await test.step('Navigate to FEDS Breadcrumbs-With-Base page', async () => {
+      await page.goto(`${baseURL}${features[0].path}?hideGeorouting=on`);
+      await page.waitForLoadState('domcontentloaded');
+      await expect(page).toHaveURL(`${baseURL}${features[0].path}?hideGeorouting=on`);
```
narcis-radu, line 17: Is this expect needed? You're going to a page and then you expect for the page to have the same URL (which is true for all use-cases).

tests/feds/breadcrumbs.test.js
```diff
@@ -0,0 +1,104 @@
...
+    await test.step('Check breadcrumbs structure on current page', async () => {
+      // Wait for FEDS GNAV to be visible:
+      await Header.mainNavContainer.waitFor({ state: 'visible', timeout: 10000 });
+      // Check breadcrumbs is displayed:
+      await expect(Header.breadcrumbContainer).toBeVisible();
+      // Check breadcrumbs structure & content:
+      await expect(Header.breadcrumbElems).toHaveCount(4);
+      let breadcrumbs = await Header.breadcrumbContainer.textContent();
+      breadcrumbs = breadcrumbs.replace(/ /g, '').replace(/(\r\n|\n|\r)/gm, '');
+      expect(breadcrumbs).toEqual('HomeProductsPhotoshopEducation');
```
narcis-radu, line 29: would we have different tests for localised content?

tests/feds/breadcrumbs.test.js
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+    await test.step('Check breadcrumbs structure on current page', async () => {
+      // Wait for FEDS GNAV to be visible:
+      await Header.mainNavContainer.waitFor({ state: 'visible', timeout: 10000 });
+      // Check breadcrumbs is displayed:
+      await expect(Header.breadcrumbContainer).toBeVisible();
+      // Check breadcrumbs structure & content:
+      await expect(Header.breadcrumbElems).toHaveCount(4);
+      let breadcrumbs = await Header.breadcrumbContainer.textContent();
```
narcis-radu, line 27: is this working? `textContent` is a property, not a method - https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent

tests/feds/breadcrumbs.test.js
```diff
@@ -0,0 +1,104 @@
...
+      breadcrumbs = breadcrumbs.replace(/ /g, '').replace(/(\r\n|\n|\r)/gm, '');
```
narcis-radu, line 52: This seems slow, here's a faster alternative https://jsperf.app/zekosi/4

tests/feds/breadcrumbs.test.js
amauch-adobe: Theoretically, the expect isn't necessarily needed, but this is a hard assertion instead of a soft assertion so if for some odd reason the page doesn't load as expected this will stop the rest of the test steps running which saves execution time in the CI/CD pipeline.  This is why we want to keep the check.

tests/feds/breadcrumbs.test.js
amauch-adobe: @narcis-radu, in this case with playwright's textContent() method on an elementHandle it will return the textContent property as you stated in your reference above.  - [https://playwright.dev/docs/api/class-locator#locator-text-content](https://playwright.dev/docs/api/class-locator#locator-text-content)

amauch-adobe (approved):
LGTM, if we can use Narcis' idea for a better performant replace method let's do it.

Dli3 (approved):

tests/feds/breadcrumbs.test.js
chadsunvice: @narcis-radu, this is pretty common QA-101 standard step & assert step. You are discounting several usual issues, like redirects, loss of browser params on route redirect, routing failures, route parsing issues, etc. In the QA world you have to always test your step was executed successfully for easier debugging purposes.

tests/feds/breadcrumbs.test.js
chadsunvice: @narcis-radu yes, and no, this is not in the scope of localized checks. We didn't start looking at localization yet from our perspective and it is out of scope of this PR.

tests/feds/breadcrumbs.test.js
narcis-radu: thanks @amauch-adobe and @chadsunvice. I have not seen the same assertions in our previous tests on PlatformUI, that's why I was curious.

narcis-radu (approved):

tests/feds/breadcrumbs.test.js
chadsunvice: It is a PlayWright locator method. Check @amauch-adobe's link.

tests/feds/breadcrumbs.test.js
chadsunvice: `breadcrumbs` is just a string with bad formatting. This seems overkill and the results when running the checks via test-runner are the same. This just over complicates things.

If you want to improve this, improve the way the breadcrumbs are formatted.

narcis-radu: The formatting is correct, this will always happen when you use `textContent` on DOM elements with children. In fact, getting `textContent` for a container with multiple elements should be avoided and instead one should loop through the children, get the text and then concatenate; this would be faster and depending on the content might not require replacing line feeds.

PR #143, opened 2023-07-17
[MWPW-133840] FEDS check improvements & ES6 support

chadsunvice:
PR relates to ticket [[NALA-Automation] Update FEDS checks (test improvements & ES6 support)](https://jira.corp.adobe.com/browse/MWPW-133840).

Added:
- support for non-US runs to not display the GeoRouting modal (impairing other page actions)
- transitioned to ES6 way of declaring classes/instantiating class objects to match Milo logic
- improved Jarvis checks (extra checks and assertions)


tests/feds/consent.test.js
```diff
@@ -14,10 +14,10 @@ test.describe('Consent Component test suite', () => {
     await test.step('Navigate to FEDS consent component page', async () => {
       // !Note: Forcing consent to load on GDPR-enforced country.
       // Load OneTrust consent component page:
-      await page.goto(`${baseURL}${features[0].path}?customPrivacyLocation=de`);
+      await page.goto(`${baseURL}${features[0].path}?customPrivacyLocation=de&hideGeorouting=on`);
```
skumar09, line 17: I think we can define it as a global scoped variable or constant and then use it, instead of changing 
it in every first `test.step()`, of a `test()` 
something like below
`const hideGeoroutingQueryParam = '?hideGeorouting=on'`;
`await page.goto(${baseURL}${features[0].path}${hideGeoroutingQueryParam});`

Also, we can think of defining a property in `.spec.js `and use it in the test.step()
`hideGeo : 'hideGeorouting=on',`
`await page.goto(${baseURL}${features[0].path}${features[0].hideGeo});`

skumar09 (changes requested):

tests/feds/jarvis.desktop.test.js
```diff
@@ -28,12 +28,17 @@ test.describe('Jarvis Component test suite', () => {
...
-      await Jarvis.jarvisButton.click();
+    await test.step('Check Jarvis component functionality', async () => {
+      // Open Jarvis chat assistent (via Jarvis button):
+      await Jarvis.openJarvisFromCta();
+      // Close Jarvis chat assistent (via Jarvis iframe):
+      await Jarvis.closeJarvisFromIframe();
+      // Open Jarvis chat assistent (via header button):
+      await Jarvis.jarvisHeaderButton.click();
       await expect(Jarvis.jarvisContainer).toBeVisible();
+      // Open Jarvis chat assistent (via header button):
```
amauch-adobe, line 39: Should this say "Close" in the comments instead of "Open" since the check is verifying it isn't visible?

amauch-adobe (changes requested):
Looks really good, just some nitpicks I found, also I agree with @skumar09 that it would be best to use a global/spec/page variable for the URL parameter to hide georouting modal since it is used across many test files.

tests/feds/jarvis.desktop.test.js
```diff
@@ -28,12 +28,17 @@ test.describe('Jarvis Component test suite', () => {
       await Consent.assertOneTrustAcceptState();
     });
 
-    await test.step('Check Jarvis component is initialized', async () => {
-      // Wait for Jarvis component to appear:
-      await Jarvis.jarvisButton.waitFor({ state: 'visible', timeout: 15000 });
-      // Open Jarvis chat assistent:
-      await Jarvis.jarvisButton.click();
+    await test.step('Check Jarvis component functionality', async () => {
+      // Open Jarvis chat assistent (via Jarvis button):
```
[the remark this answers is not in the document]
chadsunvice: Haha, yes, I agree @amauch-adobe! The Romanian in me doubled the s but wrote it as in my native language. Old-habits and no spell check in VSCode will do this. 😄

tests/feds/jarvis.desktop.test.js
```diff
@@ -58,12 +63,17 @@ test.describe('Jarvis Component test suite', () => {
...
```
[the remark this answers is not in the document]
chadsunvice: Fixed! 👍

tests/feds/jarvis.desktop.test.js
chadsunvice: Fixed! 👍

tests/feds/jarvis.desktop.test.js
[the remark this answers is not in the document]
chadsunvice: Fixed! 👍

tests/feds/consent.test.js
chadsunvice: Fixed! Added per-spec file approach since we will have different mixes of browser params which will switch per test type. We already have the consent checks which are a corner-case, so sticking to the second approach. Considered the `feds.config.js` approach, but it was more custom-coding in order to make it work.

features/feds/breadcrumbs.spec.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
       path: [
         '/libs/feds/drafts/qa/breadcrumbs/feds-breadcrumbs-with-base',
       ],
+      browserParams: '?hideGeorouting=on',
```
skumar09, line 9: It is just a perspective, you can take the call 
- from a terminology standpoint, it is referred to as a `query parameter`, you may want to consider using something like `queryParam`, 
- If it is commonly called a `browser parameter` in the applications, then `browserParams` would be easily understood by all

tests/feds/header.test.js
```diff
@@ -2,8 +2,8 @@
 /* eslint-disable no-await-in-loop */
 /* eslint-disable import/extensions */
```
skumar09, line 3: i think` /* eslint-disable import/extensions */` is not required

skumar09 (approved):
- Minor review comments, approving the PR, 
- I have added a single review comment that applies to all files. Please review and update accordingly

amauch-adobe (approved):
Approved!

features/feds/breadcrumbs.spec.js
chadsunvice: We usually call them browser params since it happens in the scope of the browser, the full naming being `URL query params`, so we default to `browser params` as a way to shorten it. The naming isn't critical as long as we know that they now reside in the `spec` files.

PR #154, opened 2023-08-24
MWPW-135502: Update page object class from named entity class to default class 

skumar09:
[MWPW-135502](https://jira.corp.adobe.com/browse/MWPW-135502): 

PR Includes

- All blocks page object classes are updated to default from named entity classes.
- Skipping Georouting test cases execution for latest-windows & webkit configuration as they are failing


chadsunvice (approved):
Looks good! 👍

JackySun9 (approved):

PR #156, opened 2023-08-24
Create dynamic test cases for UAR

JackySun9:
https://jira.corp.adobe.com/browse/MWPW-134029
https://jira.corp.adobe.com/browse/MWPW-124778

1. build test data for manual and automation
2. build test cases according to source json files
3. build automation based on dynamic test data
4. run automation side by side on prod and uar-integration


amauch-adobe (changes requested):
See my comments throughout.

libs/webutil.js
```diff
@@ -7,6 +7,7 @@ import { expect } from '@playwright/test';
 const fs = require('fs');
 // eslint-disable-next-line import/no-extraneous-dependencies
 const yaml = require('js-yaml');
+const { request } = require('@playwright/test');
```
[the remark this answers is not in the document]
JackySun9: yes, we do, it is using to load json files

features/uar/quiz.dynamic.spec.js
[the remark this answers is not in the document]
JackySun9: data may be something else later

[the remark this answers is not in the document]
JackySun9: will add @smoke and @regression later

package.json
```diff
@@ -11,7 +11,8 @@
     "test:safari": "playwright test --project=webkit --headed",
     "test:chrome-headless": "playwright test --project=chromium",
     "test:firefox-headless": "playwright test --project=firefox",
-    "test:safari-headless": "playwright test --project=webkit"
+    "test:safari-headless": "playwright test --project=webkit",
+    "test:uar": "npx playwright test --config=./configs/uar.config.js --project=uar-live-chrome --headed -g"
```
amauch-adobe, line 15: I want to be careful not to overload the package.json file in the main Nala repo.  Obviously when this project finally is made into a consumable project, then separate teams can choose to add whatever they would like in their package.json file going forward.  I remember Chris Millar wanting the package.json file to remain slim and not overloaded with different ways of executing the project (at least in this file) since it becomes difficult to read or find the execution you need.  That being said I was wondering if you couldn't use the current run.sh file to correctly send in your config and project (which I believe it is capable of doing), @skumar09 can correct me if I'm wrong.  I think the only thing that isn't currently available through the run.sh file is the '--headed' flag.  I think as a recommendation we should just add that ability to the type of flags available for executing run.sh so that we don't start having project-specific executions in the package.json file.  @skumar09, @JackySun9 , @chadsunvice , @Dli3 what do you think?

amauch-adobe (changes requested):
see my one question, but excellent job on the changes.

package.json
JackySun9: let us wait for other comments on this

skumar09: @JackySun9 : 

- Let's not add any additional test run commands in `.package.json `for consuming projects 
- Ideally, when the Nala package is consumed (in the future), you can run both UI/non-ui modes for the corresponding project.

skumar09 (changes requested):

package.json
JackySun9: ok, remove it

selectors/uar/quiz.old.page.js
```diff
@@ -0,0 +1,100 @@
...
+
+      if (name.includes('triple')) {
+        const actualProduct4 = await this.uarResult2.nth(2);
+        oldProduct.push(await actualProduct4.innerText());
+      }
+    } else {
+      oldProduct.push(text);
+    }
+
+    console.log('==========old============');
```
chadsunvice, line 96: I think we could remove this log.

selectors/uar/quiz.page.js
```diff
@@ -69,29 +67,33 @@ exports.Quiz = class Quiz {
...
+        const actualProduct4 = await this.uarResult2.nth(2);
+        newProduct.push(await actualProduct4.innerText());
       }
     } else {
-      expect(acturalProduct.toString()).toContain(expectProduct);
+      newProduct.push(text);
     }
+
+    console.log('==========new============');
+    console.log(newProduct.sort().join(''));
```
chadsunvice, line 96: Same here, combine maybe?

selectors/uar/quiz.old.page.js
```diff
@@ -0,0 +1,100 @@
...
+   * Validate products on result page to match with expect products
+   * @param {string} name
+   */
+  async checkResultPage(name) {
+    const oldProduct = [];
+
+    const actualProduct = await this.uarResult.nth(0);
+    const text = await actualProduct.innerText();
+
+    if (text.includes('We think you\'ll love')) {
```
chadsunvice, line 79: This value we check against should maybe be added in Quiz page props in case it changes in the future.

chadsunvice (approved):
Some small changes, but overall looks great @JackySun9!

skumar09 (approved):

selectors/uar/quiz.old.page.js
JackySun9: let have this here, I can update it when we need

selectors/uar/quiz.page.js
JackySun9: merged

selectors/uar/quiz.old.page.js
JackySun9: merged

amauch-adobe (approved):
Approved.

PR #161, opened 2023-09-05
MWPW-136008 : Onboarding Blog automation

skumar09:
[MWPW-136008 ](https://jira.corp.adobe.com/browse/MWPW-136008 ): Onboarding Blog automation

PR includes 
- Onboarding Blog automation to Nala
- Blog Automation test scripts for the following features 
   - Home page
   - 404 page


Dli3 (approved):

selectors/blog/404.page.js
amauch-adobe, line 1: Just a side question, but could we see this POM being a base POM that other teams could pull into their specific POM file?  Are there any class methods that would make sense in a general way that others would use?  Obviously, right now it looks like there is just a constructor, but maybe in the future.

amauch-adobe: I just know there is more than one site that uses the uncharted waters 404 page.

tests/blog/404page.test.js
amauch-adobe, line 1: Would it be good to add a test that navigates to an invalid page and verifies that the 404 page shows up since that is a functionality of the 404 page?  I like the idea of going directly to the page to test other things, but it may be worth testing that the 404 page shows up when it should.  Just an idea.

selectors/blog/home.page.js
amauch-adobe, line 1: @skumar09 I know you mentioned that we should discuss GNAV/Footer checks outside of FEDS' testing in our refinement meeting, but for some odd reason, I think there is a Base POM that @chadsunvice added.  He could add more discussion to this though.  If there is a base POM, we might want to use it here to eliminate redundancy.

amauch-adobe (approved):
Approved.

selectors/blog/home.page.js
skumar09: @amauch-adobe : My thoughts are
- Having a base or reusable fed's Gnav / Footer POM and reusable validation methods that can be used across apps test cases/steps. (The validation method should be generic and it should take navigational arguments from corresponding app specs)

cc : @Dli3

selectors/blog/404.page.js
skumar09: - This one is an initial version; it will updated soon... 
- This also qualify for base/reusable pom

PR #164, opened 2023-09-07
Add UI Screenshots for UAR

JackySun9:
https://jira.corp.adobe.com/browse/MWPW-136147
1. take UI screenshots for UAR under desktop, tablet, mobile
2. upload them to wiki for review

https://wiki.corp.adobe.com/display/WEBT/UAR+UI+Screenshots


libs/webutil.js
```diff
@@ -248,4 +248,18 @@ exports.WebUtil = class WebUtil {
   async disableNetworkLogging() {
     await this.page.unroute('**');
   }
+
+  async takeScreenshot(folderPath, desktopName, tabletName, mobileName) {
+    if (!fs.existsSync(folderPath)) {
+      fs.mkdirSync(folderPath, { recursive: true });
+    }
+    await this.page.setViewportSize({ width: 1920, height: 1080 });
```
amauch-adobe, line 256: Personal opinion here, I feel like this should be based on a config rather than hardcoded here in the webutil.js since maybe teams might have different viewport size requirements.  What does everyone else think? @skumar09 @chadsunvice @Dli3

amauch-adobe (changes requested):
See my comment about Viewport sizes.

libs/webutil.js
JackySun9: yes, we can do that, but I just have it here to make it work, if you have better idea, we can move it

amauch-adobe: A possible solution is you can add it to you uar.config.js file under the different chrome-live-desktop, safari-live-desktop, etc.

JackySun9: @amauch-adobe I will your suggestion, but not put them into config, and put them into test, for my case, I just want to take screenshot for different resolution not going to run them on different devices

amauch-adobe (approved):
Approved.

skumar09 (approved):

PR #170, opened 2023-09-18
PR branch test 

skumar09:
PR Includes
- This is work in progress test PR, it does not impact any of the existing test/code.
-
Please need aproval

### DIFF
--- a/global.setup.js
+++ b/global.setup.js
@@ -4,6 +4,11 @@ const axios = require('axios');
 
 async function globalSetup() {
   console.info('----Running Global setup---------');
+  
+  let prBranchLiveUrl;
+  let localTestLiveUrl;
+  let localOrg; 
+  let localRepo;
 
   // Check if the code is running in a GitHub CI/CD environment
   if (process.env.GITHUB_ACTIONS === 'true') {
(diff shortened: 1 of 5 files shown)


JackySun9 (approved):

global.setup.js
```diff
@@ -0,0 +1,101 @@
...
+          localRepo = match[2];
+
+          console.info('Git ORG           : ', localOrg );
+          console.info('Git REPO          : ', localRepo );
+
+          // get the current local branch name
+          const localBranch = execSync('git rev-parse --abbrev-ref HEAD', { cwd: localGitRootDir, encoding: 'utf-8' }).trim();
+          console.info('Local Branch      : ', localBranch );
+
+          if ( localRepo === 'janus' || 'nala'){
```
amauch-adobe, line 83: do we need a condition for checking if 'janus' is in the name?

amauch-adobe: I know that might be for compatibility reasons, but I don't know if there are really that many people, if any, that still have the old version of Nala when it was first named 'Janus'.

amauch-adobe (approved):
@skumar09, see my comment about the 'janus' condition statement, but other than that approved.

PR #176, opened 2023-09-20
[Modal ] Modal class attribute fix

skumar09:
PR includes a fix for the following 

- Modal class attribute update  from: ` class: 'modal link-block'` To ` class: 'modal link-block '`

### DIFF
--- a/selectors/milo/modal.block.page.js
+++ b/selectors/milo/modal.block.page.js
@@ -18,7 +18,7 @@ export default class Modal {
     // modal contents attributes
     this.attProperties = {
       'modal-link': {
-        class: 'modal link-block',
+        class: 'modal link-block ',
         'daa-ll': /^link\|(.*)/,
       },
     };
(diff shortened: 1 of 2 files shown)


JackySun9 (approved):

selectors/milo/modal.block.page.js
```diff
@@ -18,7 +18,7 @@ export default class Modal {
     // modal contents attributes
     this.attProperties = {
       'modal-link': {
-        class: 'modal link-block',
+        class: 'modal link-block ',
```
amauch-adobe, line 21: All that was added here was a space is it necessary?

amauch-adobe (changes requested):
Just curious about the extra space.

selectors/milo/modal.block.page.js
skumar09: @amauch-adobe: That was the change, (as per my earlier experience, I have seen having space, but not sure here why this is introduced...)

amauch-adobe (approved):
Approved

PR #177, opened 2023-09-21
PR branch test execution

skumar09:
PR includes

- Update local test execution url
- Clean up of print statements
- Update to playwright.config.js. for baseURL and workers count

### DIFF
--- a/global.setup.js
+++ b/global.setup.js
@@ -1,14 +1,17 @@
+import { exit } from 'process';
+
 const { execSync } = require('child_process');
 const { isBranchURLValid } = require('./libs/baseurl.js');
 const axios = require('axios');
 
 async function globalSetup() {
-  console.info('----Running Global setup---------');
+  console.info('----Executing Global setup---------');
   
   let prBranchLiveUrl;
   let localTestLiveUrl;
   let localOrg; 
   let localRepo;
+  let prNumber
 
   // Check if the code is running in a GitHub CI/CD environment
   if (process.env.GITHUB_ACTIONS === 'true') {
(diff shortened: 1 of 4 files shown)


global.setup.js
```diff
@@ -68,22 +75,22 @@ async function globalSetup() {
...
           const localBranch = execSync('git rev-parse --abbrev-ref HEAD', { cwd: localGitRootDir, encoding: 'utf-8' }).trim();
-          console.info('Local Branch      : ', localBranch );
+          console.info('Local Branch : ', localBranch );
 
-          if ( localRepo === 'janus' || 'nala'){
+          if ( localRepo === 'nala' || localRepo === 'janus'){
             localTestLiveUrl = `https://main--milo--adobecom.hlx.live`;
           } else {
-            localTestLiveUrl = `https://${localBranch}--${localRepo}--${localOrg}.hlx.live`;
+            localTestLiveUrl = `http://localhost:3000`;
```
amauch-adobe, line 93: I believe there is a possibility for another port besides just 3000.  I think there was also a proxy port (6456) that Chris Millar (you may want to ask Chris to make sure if this is still necessary) mentioned to Dennis when we were first adding the ability to run Nala against a local Franklin/Milo site.  @Dli3 created a shell script that would set up the env.js file to run on local ports, with this change I don't know if that script will be valid any longer and might be able to be removed.  Reference the script in Nala's repo top root folder, runLocalRegression.sh.

amauch-adobe (approved):
Approved. See my comment though.

global.setup.js
skumar09: @amauch-adobe : yeah that correct, 
- 6456 is running on consuming app pages with milo libraries, 
Info:  I will be enhancing the logic further in the future for accommodating running tests on consuming pages using milo libs

puruadobe (approved):

JackySun9 (approved):

PR #198, opened 2023-11-01
commenting linux-webkit project

skumar09:
PR Includes:

- Temporarily commenting Github Ubuntu - WebKit run,

### DIFF
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -58,7 +58,14 @@ const config = {
         ...devices['Desktop Firefox'],
       },
     },
-
+    /** 
+    {
+      name: 'milo-live-webkit',
+      use: {
+        ...devices['Desktop Safari'],
+      },
+    }, 
+  */
   ],
 };
 
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -59,12 +59,6 @@ const config = {
       },
     },
 
-    {
-      name: 'milo-live-webkit',
-      use: {
-        ...devices['Desktop Safari'],
-      },
-    },
   ],
 };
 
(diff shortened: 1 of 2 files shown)


playwright.config.js
```diff
@@ -59,12 +59,6 @@ const config = {
       },
     },
 
-    {
```
Dli3, line 62: If we're adding this back at a later time, can't we just comment this out for now?

Dli3 (approved):

run.sh
```diff
@@ -84,7 +84,7 @@ if [[ -n  "$APPS" ]];then
           echo "*** npx playwright test --config=./configs/${conf_name}.config.js ${TAGS} --project=${app_name} ***"
           npx playwright test --config=./configs/${conf_name}.config.js ${TAGS} --project=${app_name}-chrome ${REPORTER}
           npx playwright test --config=./configs/${conf_name}.config.js ${TAGS} --project=${app_name}-firefox ${REPORTER} 
-          npx playwright test --config=./configs/${conf_name}.config.js ${TAGS} --project=${app_name}-webkit ${REPORTER}
+          # npx playwright test --config=./configs/${conf_name}.config.js ${TAGS} --project=${app_name}-webkit ${REPORTER}
```
amauch-adobe, line 87: I might be jumping to conclusions, but if we comment this out, won't it also block other webkit runs on different OSs?

amauch-adobe (changes requested):
@skumar09 what is the issue here with the linux webkit? Also see my question below.

run.sh
skumar09: @amauch-adobe : 
- For `run-nala` label  `No` but for other labels `yes`, 
- I will revert it back in 2/3 days.

amauch-adobe: Thank you for the clarification.

amauch-adobe (approved):
Approved.

PR #213, opened 2023-11-21
Add visual test for CAAS

JackySun9:
https://jira.corp.adobe.com/browse/MWPW-138941

1. add one common method take screenshot for two urls and compare them result
2. add tests for caas team


chadsunvice (approved):

configs/visual.config.js
```diff
@@ -34,7 +34,7 @@ const config = {
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: process.env.CI
     ? [['github'], ['./utils/reporters/json-reporter.js'], ['./utils/reporters/api-reporter.js']]
-    : [['html', { outputFolder: 'test-html-results' }]],
```
skumar09, line 37: Do we need html reporter?
also please update to base-reporter from json-reporter and api-reporter, as they are not used anymore and will be removed

features/visual-compare/caas/cards.block.spec.js
```diff
@@ -0,0 +1,89 @@
...
+      tcid: '0',
+      name: '@caas-visual-onehalfcollection-desktop',
+      urlA: '/fragments/caas/test-pages/onehalfcollection',
+      urlB: '/fragments/caas/test-pages/onehalfcollection?caasver=beta',
+      tags: '@caas-visual-onehalfcollection-desktop, @caas-visual-desktop, @visual, @visual-smoke, @visual-regression',
+    },
+    {
+      tcid: '1',
+      name: '@caas-visual-threefourthscollection-desktop',
+      urlA: '/fragments/caas/test-pages/threefourthscollection',
```
skumar09, line 14: rather urlA, urlB, maybe you want to use a more intuitive 
like expected vs actual or anything better?

features/visual-compare/caas/cards.block.spec.js
```diff
@@ -0,0 +1,89 @@
...
+      urlB: '/fragments/caas/test-pages/threefourthscollection?caasver=beta',
+      tags: '@caas-visual-threefourthscollection-desktop, @caas-visual-desktop, @visual, @visual-smoke, @visual-regression',
+    },
+    {
+      tcid: '2',
+      name: '@caas-visual-halfheightcollection-desktop',
+      urlA: '/fragments/caas/test-pages/halfheightcollection',
+      urlB: '/fragments/caas/test-pages/halfheightcollection?caasver=beta',
+      tags: '@caas-visual-halfheightcollection-desktop, @caas-visual-desktop, @visual, @visual-smoke, @visual-regression',
```
skumar09, line 23: I am thinking not to use tags like test case name, i.e. `@caas-visual-halfheightcollection-desktop` as I am sure we will run either visual test or smoke or regression kind, 
- see if you can remove them, ( I am also thinking removing similar test name tags from Milo)

skumar09 (changes requested):

features/visual-compare/caas/cards.block.spec.js
JackySun9: in my case, I need this, so i will keep them

features/visual-compare/caas/cards.block.spec.js
JackySun9: update

configs/visual.config.js
JackySun9: do you test add base-reporter to any config files, it doesn't work

amauch-adobe: Are we wanting to use the generic/default visual.config.js for this test or would it be more provincial to copy this over to a specific config like we do for our other apps being tested under Nala.  I'm assuming this is for milo so maybe make a milo.visual.config.js file.

amauch-adobe (changes requested):
See my comment about the config file.

configs/visual.config.js
JackySun9: currently I am use visual.config.js, what do you think? @skumar09

JackySun9: with this config file, I just need to focus on screenshot stuff, currently one copy is good enough

JackySun9:
@amauch-adobe @skumar09 any suggestion?

amauch-adobe (approved):
Approved, but you might want to update the visual.config.js to not use the html-reporter like @skumar09 mentioned since we no longer use that one.

JackySun9:
@skumar09 @amauch-adobe I updated reporter in visual.config.js according to santosh's guide, we should be good now

PR #216, opened 2023-11-29
add screenshot for UAR stable and beta

JackySun9:
1. use milo main live as stable
2. use uar-integration as beta
3. run on the same content and take screenshots
4. compare the screenshots, save diff images if exist
5. add a script to record time stampe


amauch-adobe (approved):
Approved.

PR #239, opened 2024-01-10
Update final path for UAR in CC

JackySun9:
The final path should be /creativecloud/plan-recommender/


skumar09 (approved):
Looks good....

PR #262, opened 2024-02-07
Adding marketo forms tests for the 6 different templates

Dli3:
- Adding Marketo form tests for the 6 different Marketo templates.
- Removed the test for short form since it's not going to be used.
- Updated company locator which was causing failures in daily runs.

Note: 

- There's currently an issue on Firefox where you're unable to enter input in the company field after selecting a country. And selecting a country will remove the company input if it was input first. I'll look into this.


amauch-adobe (approved):
Approved, it looks like the test is also failing in chrome as shown in the github action execution.  I would check that out before having this merged.

JackySun9 (approved):

Dli3:
Thanks @amauch-adobe . The failing tests are no longer related to the marketo forms tests. Merging pr.

PR #267, opened 2024-02-14
Helpx_Viewport Specific Container cases

kirupacommit:
vps


kirupacommit:
@skumar09 @amauch-adobe  Kindly review above PR

tests/helpx/vps.test.js
amauch-adobe, line 1: Just two comments, throughout the test file there was viewport config changes (which is totally fine), but I wanted to point out you can do the same thing for a project config and name it specific to the type of viewport size that will run all tests in that specific viewport.  That being said you may have only wanted this test to be run this way with this viewport without having to send in specific tags for execution to use the project config and filter it down to just this test.

Second comment, is always be careful about the usage of waitForTimeout as stated in playwrights documentation here (https://playwright.dev/docs/api/class-page#page-wait-for-timeout), it is a general recommendation to not use it in production or to waitFor a specific event or expected behavior with an element, or to use locators to happen before running the assertion expectation.  Timeout waits may cause false failures/flakiness due to VM starts, network situations, etc. in CICD pipelines.

amauch-adobe (approved):
Approve.

selectors/helpx/vps.page.js
```diff
@@ -0,0 +1,34 @@
+export default class vps {
+  constructor(page) {
+    this.page = page;
+
+    // vps Selectors for hidden desktop
+    this.vpsForHiddenDesktop = page.locator("//div[contains(@daa-lh,'b2|generic')][contains(@class,'generic hidden-desktop')]");
```
Dli3, line 6: You could chain the locators.
https://playwright.dev/docs/locators#matching-inside-a-locator   
For example: 
`this.vpsForImginHiddenDesktop = this. vpsForHiddenDesktop .locator("/descendant::picture/img")`

tests/helpx/vps.test.js
```diff
@@ -0,0 +1,344 @@
...
+  //create a new page
+  page = await context.newPage();
+  vps = new vpsPage(page); 
+  
+});
+
+test.describe('Verify hidden Tablet and mobile and Desktop for video in Viewport specific container', () => {
+
+  test(`${features[7].name}, ${features[7].tags}`, async ({ baseURL }) => {
+    let vpsTag = features[7].path;
```
Dli3, line 35: vpsTag doesn't seem to be changing within your test. 
```suggestion
    const vpsTag = features[7].path;
```

tests/helpx/vps.test.js
```diff
@@ -0,0 +1,344 @@
...
+          await expect(vps.linkForHiddenTabletMobileB26).not.toBeVisible();
+
+          // Wait for 1000 milliseconds
+          await page.waitForTimeout(1000);
+
+          // Set viewport size for tablet
+          await page.setViewportSize({ width: 768, height: 1024 }); // iPad viewport size
+          await expect(vps.vpsForHiddenTabletMobileB26).not.toBeVisible();
+          await expect(vps.linkForHiddenTabletMobileB26).not.toBeVisible();
+
```
Dli3, line 62: Do we need these empty lines?

tests/helpx/vps.test.js
```diff
@@ -0,0 +1,344 @@
...
+test.describe('Verify hidden desktop in Viewport specific container', () => {
+
+  test(`${features[0].name}, ${features[0].tags}`, async ({ baseURL }) => {
+    let vpsTag = features[0].path;
+    await page.goto(`${helpxconfig.use?.baseURL}${vpsTag}`);
+    await page.waitForLoadState('domcontentloaded');
+    console.log(`[Test Page]: ${baseURL}${vpsTag}`);
+    //I verify url of the page
+    await test.step('Navigate to Viewport specific container page', async () => {
+      await expect(page).toHaveURL(`${baseURL}${vpsTag}`);
```
Dli3, line 77: Do you need the `${helpxconfig.use?.`part that you have above?

tests/helpx/vps.test.js
```diff
@@ -0,0 +1,344 @@
...
+});
+
+test.describe('Verify hidden desktop in Viewport specific container', () => {
+
+  test(`${features[0].name}, ${features[0].tags}`, async ({ baseURL }) => {
+    let vpsTag = features[0].path;
+    await page.goto(`${helpxconfig.use?.baseURL}${vpsTag}`);
+    await page.waitForLoadState('domcontentloaded');
+    console.log(`[Test Page]: ${baseURL}${vpsTag}`);
+    //I verify url of the page
```
Dli3, line 75: Could be a part of your test.step name.

tests/helpx/vps.test.js
```diff
@@ -0,0 +1,344 @@
...
+
+    await page.setViewportSize({ width: 1280, height: 800 }); // Typical desktop viewport size
+
+    //Hidden Desktop
+    await expect(vps.vpsForHiddenTabletB4).toBeVisible();
+    await expect(vps.vpsForImginHiddenTabletB4).toBeVisible();
+
+       await page.waitForTimeout(1000);
+      // Set viewport size for mobile
+      await page.setViewportSize({ width: 375, height: 812 }); // iPhone X viewport size
```
Dli3, line 125: You have two very similar comments. Maybe remove one of them?

tests/helpx/vps.test.js
kirupacommit: Removed

tests/helpx/vps.test.js
kirupacommit: Done. Thanks For your suggestion

tests/helpx/vps.test.js
kirupacommit: Thank you for your insight, removed ${helpxconfig.use?}

kirupacommit:
@skumar09 can u please review my PR with updated changes

PR #278, opened 2024-03-01
Browserstack integration

Dli3:
Ticket: [MWPW-122014](https://jira.corp.adobe.com/browse/MWPW-122014) 

- Adding Browserstack config so we can run tests against the available device and OS options in browserstack through their Browserstack SDK ([Their recommended approach as the other option is now legacy](https://www.browserstack.com/docs/automate/playwright/getting-started/nodejs/integrate-your-tests-sdk))
- Updated playwright version to 1.38.1. As per their support, we needed playwright version 1.38.1 or higher. 
- [iOS and Android isn't supported at this time.](https://www.browserstack.com/docs/automate/playwright/browsers-and-os) Will add those when the option is available. The documentation specifies that version 1.38.1 has Android chrome as an option but I couldn't get it to work. I've reached out to support and will update the yml file when I can get it working.

How to run your test in Browserstack:
1. Install the dependencies. 
2. Set up your BROWSERSTACK_USERNAME and BROWSERSTACK_ACCESS_KEY env variables.
3. Command: `npx browserstack-node-sdk playwright test --config='./configs/<project>.config.js'`.

### DIFF
--- a/browserstack.yml
+++ b/browserstack.yml
@@ -46,7 +46,7 @@ platforms:
   - os: OS X
     osVersion: Sonoma
     browserName: Chrome
-    browserVersion: latest
+    browserVersion: 121.0
   - os: OS X
     osVersion: Sonoma
     browserName: playwright-firefox
--- a/browserstack.yml
+++ b/browserstack.yml
@@ -40,32 +40,16 @@ platforms:
     browserName: playwright-firefox
     browserVersion: latest
   - os: OS X
-    osVersion: Monterey
+    osVersion: Sonoma
+    browserName: playwright-webkit
+    browserVersion: latest
+  - os: OS X
+    osVersion: Sonoma
     browserName: Chrome
     browserVersion: latest
   - os: OS X
-    osVersion: Ventura
-    browserName: Chrome
-    browserVersion: latest
-  - os: OS X
-    osVersion: Ventura
+    osVersion: Sonoma
     browserName: playwright-firefox
-    browserVersion: latest
-  - os: OS X
-    osVersion: Monterey
-    browserName: playwright-webkit
-    browserVersion: latest
-  - os: OS X
-    osVersion: Ventura
-    browserName: playwright-webkit
-    browserVersion: latest
-  - os: OS X
-    osVersion: Sonoma
-    browserName: playwright-webkit
-    browserVersion: latest
-  - os: OS X
-    osVersion: Sonoma
-    browserName: Chrome
     browserVersion: latest
 
 # =======================
@@ -85,7 +69,7 @@ parallelsPerPlatform: 1
 # ==========================================
 # Set browserStackLocal to true if your website under test is not accessible publicly over the internet
 # Learn more about how BrowserStack Local works here -> https://www.browserstack.com/docs/automate/selenium/local-testing-introduction
-browserstackLocal: true # <boolean> (Default false)
+browserstackLocal: false # <boolean> (Default false)
 # browserStackLocalOptions:
 # Options to be passed to BrowserStack local in-case of advanced configurations
 # localIdentifier: # <string> (Default: null) Needed if you need to run multiple instances of local.
--- a/browserstack.yml
+++ b/browserstack.yml
@@ -1,0 +1,106 @@
+# =============================
+# Set BrowserStack Credentials
+# =============================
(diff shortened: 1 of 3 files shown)


JackySun9:
did you run all the tests for this?

kirupacommit:
Hi jacky,

My Test cases are failing due to environment variable changes. So to debug
the issue , I ran the test cases.

May be i will run one more time if fix is merged

Dli3:
> did you run all the tests for this?

Yes, I ran all the tests on the bstack branch vs main. I didn't see any issues with the update. There are failures with some of the tests, but I confirmed that it was happening on the main branch as well.

JackySun9 (approved):

browserstack.yml
```diff
@@ -0,0 +1,106 @@
...
+    browserName: playwright-firefox
+    browserVersion: latest
+  - os: OS X
+    osVersion: Monterey
+    browserName: Chrome
+    browserVersion: latest
+  - os: OS X
+    osVersion: Ventura
+    browserName: Chrome
+    browserVersion: latest
```
amauch-adobe, line 49: 2 quick questions. 

1. Is there a reason to have Monterey, Ventura, and Sonoma instead of just the latest Mac OSX listed since my guess is this is a template for teams to adjust as needed for their own needs?  
2. I noticed Firefox doesn't get configs for Sonoma or Monterey, is there a reason for this?

browserstack.yml
```diff
@@ -0,0 +1,106 @@
...
+# Example 2 - If you have configured 1 platform and set `parallelsPerPlatform` as 5, a total of 5 (1 * 5) parallel threads will be used on BrowserStack
+parallelsPerPlatform: 1
+
+# ==========================================
+# BrowserStack Local
+# (For localhost, staging/private websites)
+# ==========================================
+# Set browserStackLocal to true if your website under test is not accessible publicly over the internet
+# Learn more about how BrowserStack Local works here -> https://www.browserstack.com/docs/automate/selenium/local-testing-introduction
+browserstackLocal: true # <boolean> (Default false)
```
amauch-adobe, line 88: would it be better to leave this value "browserstackLocal" as false?  The reason I say that is if this is to be used as a template for teams it would make more sense to have values at their defaults unless it is absolutely necessary for all executions/runs/setups regardless of project being tested.

package.json
```diff
@@ -13,7 +13,19 @@
     "test:firefox-headless": "playwright test --project=firefox",
     "test:safari-headless": "playwright test --project=webkit",
     "test.sauce.us": "saucectl run --region us-west-1",
-    "test.sauce.eu": "saucectl run"
+    "test.sauce.eu": "saucectl run",
+    "postinstall": "npm update browserstack-node-sdk",
+    "test:browserstack": "npx browserstack-node-sdk playwright test",
+    "test:bstack-acom": "npx browserstack-node-sdk playwright test --config='./configs/acom.config.js'",
```
amauch-adobe, line 19: Its entirely up to you, this is just a personal opinion, but if --config can be something sent in as a parameter on the command line execution, I feel it would be best to limit the named npm execution commands to just the bare minimum to run it.  Then users can just add the config parameter themselves and the package.json file is cleaner and not so large with multiple command paths.

amauch-adobe (changes requested):
See my questions, comments.  Looks great Dennis, excellent job.

browserstack.yml
Dli3: Thanks I changed it when I was testing this and missed changing it back. I updated it back to false.

package.json
Dli3: > Its entirely up to you, this is just a personal opinion, but if --config can be something sent in as a parameter on the command line execution, I feel it would be best to limit the named npm execution commands to just the bare minimum to run it. Then users can just add the config parameter themselves and the package.json file is cleaner and not so large with multiple command paths.

That's a good point. I've removed the added lines and left test:bstack.

browserstack.yml
Dli3: I thought it would be useful to add Monterey and Ventura when Safari becomes available since we test two versions back. But since that's not available yet, I've removed the other versions and kept Ventura. 

I've added Firefox on Mac Ventura. 

Thanks!

amauch-adobe (approved):
Approved.

Dli3:
Closing PR to fix a workflow issue. I'll open another PR.

PR #281, opened 2024-03-02
setup automation for cc uar quiz

JackySun9:
1. test uar live page for CC


features/cc/quiz.dynamic.spec.js
```diff
@@ -0,0 +1,194 @@
+module.exports = {
+  name: 'Quiz Recommender Blocks',
+  features: [
+    {
+      tcid: '10',
```
amauch-adobe, line 5: not that it matters too much, but why are the tcid numbers starting so high and then jumping to the 20's and 30's?  Just wondering if there was a reason for it or if it should start at the normal 0 and forward numbers.

features/cc/quiz.spec.js
```diff
@@ -0,0 +1,75 @@
+module.exports = {
+  name: 'Quiz Recommender Blocks',
+  features: [
+    {
+      tcid: '000',
```
amauch-adobe, line 5: Is there a reason for the tcid being '000' and not just '0' and then the rest of the tcids in this file following a normal increment?

tests/cc/quiz.dynamic.test.js
amauch-adobe, line 1: Lint error should be resolved.

tests/uar/quiz.screenshots.test.js
amauch-adobe, line 1: Lint error should be resolved.

amauch-adobe (changes requested):
See my comments and questions on the spec.js files.  Also, it would be best to have the Linter errors resolved before being merged.

features/cc/quiz.dynamic.spec.js
JackySun9: I use id to category them

features/cc/quiz.spec.js
JackySun9: ok, I can update this

amauch-adobe (approved):
Approved.

PR #289, opened 2024-03-09
Browserstack Integration

Dli3:
Adding Browserstack config and updating playwright version to 1.41.2.
Ticket: [MWPW-122014](https://jira.corp.adobe.com/browse/MWPW-122014)

Adding Browserstack config so we can run tests against the available device and OS options in browserstack through their Browserstack SDK ([Their recommended approach as the other option is now legacy](https://www.browserstack.com/docs/automate/playwright/getting-started/nodejs/integrate-your-tests-sdk))
Updated playwright version to 1.41.2. As per their support, we needed playwright version 1.38.1 or higher.
[iOS and Android isn't supported at this time.](https://www.browserstack.com/docs/automate/playwright/browsers-and-os) Will add those when the option is available. The documentation specifies that version 1.41.2 has Android chrome as an option but I couldn't get it to work. I've reached out to support and will update the yml file when I can get it working.
How to run your test in Browserstack:

Install the dependencies.
Set up your BROWSERSTACK_USERNAME and BROWSERSTACK_ACCESS_KEY env variables.
Command: npx browserstack-node-sdk playwright test --config='./configs/<project>.config.js'.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,3 +132,8 @@ Test.pptx
 nala-results.json
 configs/screenshots/
 configs/test-html-results/
+
+# Browserstack files
+browserstackSetupConfig.json
+playwright-browserstack-sdk.config.js
+playwright-browserstack-sdk.config.temp.json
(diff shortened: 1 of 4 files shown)


Dli3:
We'll upgrade our playwright version first then add browserstack.

PR #298, opened 2024-03-22
Setup API automation to validate ML results

JackySun9:
MWPW-142970: Setup API automation to validate ML results


package.json
```diff
@@ -15,7 +15,11 @@
     "test.sauce.us": "saucectl run --region us-west-1",
     "test.sauce.eu": "saucectl run",
     "postinstall": "npm update browserstack-node-sdk",
-    "test:bstack": "npx browserstack-node-sdk playwright test"
+    "test:bstack": "npx browserstack-node-sdk playwright test",
+    "test:uar": "npx playwright test --config=./configs/uar.config.js --project=uar-live-chrome --headed -g",
```
amauch-adobe, line 19: Chris Millar requested us a while back with the Nala package.json file to not have so many execution commands but instead to keep it more concise and simple.  If there is a way to limit how many commands we have and instead look to having npm commands that add options via the command line to keep the package.json file cleaner/leaner, I feel it would be best.

tests/uar/quiz.ai.test.js
```diff
@@ -0,0 +1,83 @@
...
+/* eslint-disable no-restricted-syntax */
+import { test } from '@playwright/test';
+
+const QuizSpec = require('../../features/uar/quiz.ai.spec.js');
+
+const { features } = QuizSpec;
+const { WebUtil } = require('../../libs/webutil.js');
+
+const ENDPOINT = 'acom-prd-recom-v3';
+const API_KEY = 'CCHomeMLRepo1';
```
amauch-adobe, line 12: will want to have this be a github secret.

amauch-adobe (changes requested):
See my comments below.  Excellent job overall.

package.json
JackySun9: ok, I can clean up this

tests/uar/quiz.ai.test.js
JackySun9: This is ok, it is client id, not private key or secret

PR #301, opened 2024-03-28
Enabing milo tests on bacom

Dli3:
[MWPW-145790](https://jira.corp.adobe.com/browse/MWPW-145790)
Enabing milo tests on bacom.

### DIFF
--- a/configs/bacom.config.js
+++ b/configs/bacom.config.js
@@ -8,7 +8,7 @@ const envs = require('../envs/envs.js');
  * @type {import('@playwright/test').PlaywrightTestConfig}
  */
 const config = {
-  testDir: '../tests/bacom',
+  testDir: '../tests',
   outputDir: '../test-results',
   /* Maximum time one test can run for. */
   timeout: 45 * 1000,


JackySun9 (approved):

Dli3:
> Approved. Although, I think it would be good to look at the sustainability of this approach since you will need to keep updating your sharepoint directory with any changes made for milo core tests to ensure they are in sync with one another. I'm guessing this ticket was derived from the desire to make sure any BAcom code changes didn't affect Milo core blocks prior to it going to staging and being merged into Milo (basically proactively ensuring code changes don't break Milo core libs/blocks). It might be prudent to discuss these needs in the next Nala Sync meeting since other teams may want to do this and might require a more sustainable approach.
> 
> cc @JackySun9 and @skumar09 can chime in.

This is to help bacom SOT with stage to main verification. But I agree that there will be a bit of maintenance work with updating the sharepoint files. 

We can discuss in the sync meeting. 
cc @JackySun9 @skumar09

PR #311, opened 2024-04-04
update id to be unique

JackySun9:


package.json
```diff
@@ -15,7 +15,8 @@
     "test.sauce.us": "saucectl run --region us-west-1",
     "test.sauce.eu": "saucectl run",
     "postinstall": "npm update browserstack-node-sdk",
-    "test:bstack": "npx browserstack-node-sdk playwright test"
+    "test:bstack": "npx browserstack-node-sdk playwright test",
+    "test:uar": "npx playwright test --config=configs/uar.config.js --project=uar-live-chrome --headed -g"
```
Dli3, line 19: I don't think we want to add team specific commands here.

skumar09: I agree with @Dli3 , let's not have consuming application specific command

features/uar/quiz.spec.js
```diff
@@ -2,74 +2,11 @@ module.exports = {
   name: 'Quiz Recommender Blocks',
   features: [
     {
-      tcid: '000',
+      tcid: '0',
       name: '@quiz basic',
       path: '/drafts/quiz/quiz-2/',
-      tags: '@cc @uar-quiz-basic @uar-quiz-static',
+      tags: '@cc @uar-quiz-basic-milo @uar-quiz-static',
```
skumar09, line 8: @JackySun9 : You may want to generalize these tags ? as they are long string [ `@uar-quiz-basic-milo`  ]

skumar09 (approved):
- PR Looks good

- Have few minor comments, please review them

PR #329, opened 2024-04-18
Added tests for the new marketo forms templates

Dli3:
[MWPW-146473](https://jira.corp.adobe.com/browse/MWPW-146473)
Added tests for the new marketo form templates.

### DIFF
--- a/features/milo/marketo.block.spec.js
+++ b/features/milo/marketo.block.spec.js
@@ -2,39 +2,64 @@ module.exports = {
   name: 'Marketo Forms block',
   features: [
     {
+      tcid: '0',
       name: '@marketo production form',
       path: '/drafts/nala/blocks/marketo/marketo-mcz-production-2277',
       tags: '@marketo @marketoProductionForm @milo @smoke @regression',
     },
     {
+      tcid: '1',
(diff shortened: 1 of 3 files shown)


features/milo/marketo.block.spec.js
```diff
@@ -5,36 +5,61 @@ module.exports = {
       name: '@marketo production form',
       path: '/drafts/nala/blocks/marketo/marketo-mcz-production-2277',
       tags: '@marketo @marketoProductionForm @milo @smoke @regression',
+      tcid: '0',
```
JackySun9, line 8: move this to the first one

Dli3: I believe we start with 0 (See the other milo tests). If we changed the first test case id to 1, it'll be out of sync with the tag index used in the test. @JackySun9

JackySun9: sorry, move to the top, first line

Dli3: Ok, I've moved the tcid lines. @JackySun9

JackySun9 (approved):

PR #335, opened 2024-04-23
MWPW-146964: [Nala] unified test run commands 

JackySun9:
1. support run on all projects with -p 'all'
2. support run some projects with comma-separated list of projects, such as milo-live-chrome,milo-live-firefox


amauch-adobe (approved):
Approved.

PR #343, opened 2024-05-02
Adding link comparison script.

Dli3:
[MWPW-146942](https://jira.corp.adobe.com/browse/MWPW-146942)
Adding link comparison script.

This script will go in and grab the links on page A and B. Then check that page B has the same hrefs associated to the link texts found in page A.


tests/bacom-blog/link-comparison.test.js
```diff
@@ -0,0 +1,61 @@
+import fs from 'fs';
+import { test } from '@playwright/test';
+import { findLinks, checkForDiff } from '../../utils/linkComparison.js';
+
+const data = JSON.parse(fs.readFileSync('./data/bacom-blog/stagedBlogUrls.json', 'utf8'));
```
JackySun9, line 5: you don't include the json file here

tests/bacom-blog/link-comparison.test.js
```diff
@@ -0,0 +1,61 @@
...
+        if (res.status() === 404) {
+          results[urlB] = 404;
+          test.skip();
+        }
+
+        await page.waitForLoadState('domcontentloaded');
+        linksB = await findLinks(page);
+      });
+
+      await test.step('3. Verify that the links are the same in url A and B', async () => {
```
JackySun9, line 55: how do you make sure, urlA is comparing urlB, I mean if there is more links in urlB, the links will not match

JackySun9 (changes requested):

tests/bacom-blog/link-comparison.test.js
Dli3: @JackySun9 We just want to make sure the links from A were not shuffled in B. There could've been authoring changes to B at the time of running the script, so we're not checking if they have the same length.

tests/bacom-blog/link-comparison.test.js
Dli3: Do we want to include data files in our repo? @JackySun9 
cc @skumar09 @amauch-adobe

JackySun9: are the urls dynamic?

tests/bacom-blog/link-comparison.test.js
JackySun9: could you give some examples?

utils/linkComparison.js
```diff
@@ -0,0 +1,45 @@
...
+
+  Object.entries(a).forEach(([key, value]) => {
+    const aValue = value ? value.toString() : value;
+    const bValue = b[key] ? b[key].toString() : b[key];
+    if (aValue !== bValue) {
+      diff[key] = [aValue, bValue];
+    }
+  });
+
+  return diff;
```
JasonHowellSlavin, line 44: Could use reduce, no need to change. 

```suggestion
export async function checkForDiff(a, b) {
  if (Object.keys(b).length === 0) return {};

  return Object.entries(a).reduce((rdx, [key, value]) => {
    const aValue = value ? value.toString() : value;
    const bValue = b[key] ? b[key].toString() : b[key];
    if (aValue !== bValue) {
      rdx[key] = [aValue, bValue];
    }
    return rdx;
  }, {});
```

JasonHowellSlavin (commented):
Looks good, nothing needs to change, but some suggestions.

tests/bacom-blog/link-comparison.test.js
amauch-adobe: @JackySun9 the reason for this test is to address a CSO issue where between bulk updating blog files the urls were no longer the same as they were before, meaning the links swapped around in the file that caused the links to no longer be valid.  It has led to discussions about the Franklin process of the word doc in sharepoint being converted to MarkDown for Franklin to use and then converted back to a Word doc with the updates in Sharepoint being there being the culprit for the scrambling of the links (according to Microsoft support).  This test is to help that situation.

utils/linkComparison.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * @description Finds the main tag and returns all the a tags found.
```
amauch-adobe, line 2: might want to change the 'a tags' to '<a> tags' in this comment so it is more clear.

amauch-adobe (approved):
Approved, made a small nit pick grammar change in a comment, but otherwise approved.

utils/linkComparison.js
Dli3: Updated. Thanks @JasonHowellSlavin .

tests/bacom-blog/link-comparison.test.js
Dli3: I didn't plan for it to be ran in CI/CD. @amauch-adobe 

@JackySun9 I've added the json file.

tests/bacom-blog/link-comparison.test.js
```diff
@@ -0,0 +1,61 @@
...
+        const res = await page.goto(urlB);
+
+        if (res.status() === 404) {
+          results[urlB] = 404;
+          test.skip();
+        }
+
+        await page.waitForLoadState('domcontentloaded');
+        linksB = await findLinks(page);
+      });
```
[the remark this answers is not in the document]
Dli3: Thanks for the suggestion. I've updated it. @JasonHowellSlavin

skumar09 (approved):
@Dli3 : This looks good, as discussed will raise enhancement pr for link difference reporting, (like csv, or html)

JackySun9 (approved):

PR #349, opened 2024-05-21
MWPW-147876 Improve SharePoint editing

Brandon32:
* Add browser context for headless runs, login once with headed.
* Fetch admin api instead of navigating to save time.
* Check that document loads successfully and output any errors.
* Navigate to first non-table section to be able to add page break, checks for page break.
* Soft fail if document doesn't save, undo edit step.
* Tuning timeouts: 1s for `toPass` loops, 30s for long tasks, 2 min timeout for each test. (Most pages finish in under 20s.)

Resolves: [MWPW-147876](https://jira.corp.adobe.com/browse/MWPW-147876)

### DIFF
--- a/selectors/bacom-blog/sharepoint.page.js
+++ b/selectors/bacom-blog/sharepoint.page.js
@@ -11,39 +11,87 @@ export default class Sharepoint {
     this.insertPageBreakButton = this.iframe.locator('#InsertPageBreak');
     this.homeButton = this.iframe.locator('#Home');
     this.undoButton = this.iframe.locator('#UndoRedo > button:first-of-type');
+    this.editingElement = this.iframe.locator('#WACViewPanel_EditingElement');
+    this.documentTitle = this.iframe.locator('button#documentTitle');
+    this.dialogPanel = this.iframe.locator('#WACDialogPanel');
+    this.dialogText = this.iframe.locator('#WACDialogTextPanel');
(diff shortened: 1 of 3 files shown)


selectors/bacom-blog/sharepoint.page.js
```diff
@@ -11,39 +11,87 @@ export default class Sharepoint {
...
 
   /**
-   * @description Adds a page break by selecting the insert page break button under
-   * the inserts tab and then saves the file.
+   * @description Gets the text of the dialog box.
+   * @returns {Promise<string>} The text of the dialog box.
+   */
+  async getDialogText() {
+    if (await this.dialogText.isVisible()) {
+      return this.dialogText.textContent();
```
JackySun9, line 28: you need await for this

JackySun9: I mean for textContent()

Dli3 (approved):

selectors/bacom-blog/sharepoint.page.js
Brandon32: ESLint will complain of redundant use of `await` on a return. (https://eslint.org/docs/rules/no-return-await)
I can change it to this:
```suggestion
      const dialogText = await this.dialogText.textContent();
      return dialogText;
```

JackySun9 (approved):

JasonHowellSlavin (approved):

PR #365, opened 2024-06-05
Minor OST Price selectors update as per latest product changes

sigadamvenkata:
All cases passed , below screen shot
image


JackySun9 (approved):

JackySun9:
if we don't use text in locator, that will be perfect

PR #383, opened 2024-06-21
MWPW-153095 Capture Sharepoint Session ID

Brandon32:
* Capture Session ID to share with Microsoft
* Output results to JSON file for future session lookup

Resolves: [MWPW-153095](https://jira.corp.adobe.com/browse/MWPW-153095)


JasonHowellSlavin (approved):
LGTM, though did we want to commit the `data/bacom/stagedUrls` file ?

Dli3 (approved):

Brandon32:
> LGTM, though did we want to commit the `data/bacom/stagedUrls` file ?

I think we should have some entries so anyone could run it, this is the first batch but I could commit a shorter list.

PR #411, opened 2024-07-26
add dc screenshots

JackySun9:
add dc sot screenshots


Dli3 (approved):

tedAir (commented):
Can we make the wait time for Beta longer? Like Double the amount? Page https://www.stage.adobe.com/acrobat/online/word-to-pdf.html loads much longer.

PR #431, opened 2024-09-06
update video tests

JackySun9:


Dli3 (approved):

amauch-adobe (approved):
Approved, based on the request from Milo Core to remove the no-longer needed checks on attributes, but instead to verify only visibility.

skumar09 (approved):

PR #453, opened 2024-10-15
Add 404 checking for graybox-bacom

JackySun9:
Add 404 checking for graybox-bacom


amauch-adobe (approved):
Approved.  I would just make sure to have in the .gitignore file the 404 results .json file so that it doesn't get committed to the repo, unless of course you have a reason for committing it.

configs/graybox-bacom.config.js
```diff
@@ -9,7 +9,7 @@ const envs = require('../envs/envs.js');
  */
 const config = {
   testDir: '../tests/',
-  testMatch: ['bacom/**/*.test.js', 'milo/**/*.test.js'],
+  testMatch: ['bacom/**/*.test.js', 'milo/**/*.test.js', 'graybox/**/*.test.js'],
```
skumar09, line 12: @JackySun9 : you want to include milo tests runs also

skumar09 (approved):

configs/graybox-bacom.config.js
JackySun9: yes, we need to run milo test for graybox webistes

JackySun9:
> Approved. I would just make sure to have in the .gitignore file the 404 results .json file so that it doesn't get committed to the repo, unless of course you have a reason for committing it.

yes, fixed, thanks!

PR #462, opened 2024-11-07
Cards

hadobe:


skumar09 (approved):

configs/express.config.js
```diff
@@ -58,6 +58,14 @@ const config = {
         baseURL: envs['@express_live'],
       },
     },
+
+    {
+      name: 'express-live-webkit',
+      use: {
+        ...devices['Desktop Safari'],
+        baseURL: envs['@express_live'],
```
JackySun9, line 66: if you run this on express live with automation, might cause hight traffic, do you have stage server

hadobe: Is it true for Firefox and Chrome as well?

JackySun9: yes, here you are using express_live, which is pointing to express hlx live, right?

hadobe: yes, example - https://main--express--adobecom.hlx.live/drafts/nala/blocks/cards-dark

So you suggest I point to stage?

JackySun9: Please check with your team to get suggestion?

JackySun9: I just approve this now

JackySun9 (approved):

PR #463, opened 2024-11-14
update screenshot tests to public s3

JackySun9:
1. update screenshot tests to public s3
2. add git action to run them


.github/workflows/run-screendiff.yml
```diff
@@ -0,0 +1,50 @@
...
+  test-matrix:
+    name: Running ${{ inputs.category }} tests
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: windows-latest
+            categories: ["sot","milo","feds","caas","uar","uar-ai"]
+          - platform: macos-latest
+            categories: ["dc","bacom","graybox-bacom","graybox-dc"]
```
amauch-adobe, line 21: This is just a question, not necessarily an issue.  Is there a reason for why categories only run on a certain platform and not both?

amauch-adobe (approved):
Had a comment/question, but otherwise approved.

PR #466, opened 2024-12-05
fix graybox-cc screenshot tests

JackySun9:
fix graybox-cc screenshot tests


tests/visual/graybox/sot.cc.test.js
```diff
@@ -12,32 +12,43 @@ const results = {};
...
-        );
-        results[name] = [result];
+        try {
+          // Add retry logic and longer timeout for navigation
+          const result = await takeTwo(
+            page,
+            stableURL,
+            async () => {
+              await page.waitForLoadState('networkidle');
+              await page.waitForTimeout(8000);
```
amauch-adobe, line 32: how arbitrary are these timeouts? Is there a better way for waiting on the page to fully load, like waiting for a specific element in the last section/block to load completely.  Just wondering since there may be a potential for these timeouts to fluctuate.

tests/visual/graybox/sot.cc.test.js
```diff
@@ -12,32 +12,43 @@ const results = {};
...
+            },
+            betaURL,
+            async () => {
+              await page.waitForLoadState('networkidle');
+              await page.waitForTimeout(10000);
```
amauch-adobe, line 37: Same as above comment.

amauch-adobe (approved):
Approved, but I did leave a couple comments/questions to consider that may or may not be applicable.  Just something to consider.

tests/visual/graybox/sot.cc.test.js
amauch-adobe: If it works, just a thought, you could make a parameter that allows you to send in an element to look for or by default if nothing is sent in use the arbitrary timeouts.  Not a requirement, but could potentially help if the tests become flaky due to the timeouts.

PR #482, opened 2025-02-26
Adding urls for the headers tests

Dli3:
[MWPW-168626](https://jira.corp.adobe.com/browse/MWPW-168626) 
Adding request consultation test pages to the headers tests.


features/bacom/headers.spec.js
```diff
@@ -12,6 +12,20 @@ module.exports = {
...
+        '/fr/request-consultation/experience-cloud.html',
+        '/uk/request-consultation/experience-cloud.html',
+        '/request-consultation.html',
+        '/de/request-consultation.html',
+        '/fr/request-consultation.html',
+        '/au/request-consultation.html',
+        '/uk/request-consultation.html',
+        '/request-consultation/thankyou.html',
+        '/uk/request-consultation/thankyou.html',
+        '/de/request-consultation/thankyou.html',
```
amauch-adobe, line 28: is their not locale specific thank you pages for 'fr' and 'au'? I see the form pages in the list, just not their thankyou.html pages.  So I was just wondering.

amauch-adobe (approved):
Approved.

features/bacom/headers.spec.js
Dli3: I picked a few pages to test the form and thank you pages. Figured it'll be sufficient and that'll confirm that the akamai rule is working. 

Do you think I should add the all the form and thank you pages in all tier 1 locale? @amauch-adobe

amauch-adobe: I believe these all use the same form just translated, I think your fine.  Now that I think of it, the thank you page doesn't need to be tested across all tiers, that does seem unnecessary.  It was just first glance notice that it wasn't the same one-to-one.  You are good!

PR #489, opened 2025-03-21
Adding Marketo error messages check

Dli3:
[MWPW-142682 ](https://jira.corp.adobe.com/browse/MWPW-142682)
Adding Marketo error messages check.

### DIFF
--- a/.github/workflows/run-lint.yml
+++ b/.github/workflows/run-lint.yml
@@ -20,6 +20,6 @@ jobs:
         run: npm ci
 
       - name: Run eslint on changed files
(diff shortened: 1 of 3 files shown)


amauch-adobe (approved):
Approved.  LGTM

amauch-adobe:
It looks like the Lint job failed, but the error message looks like the job itself had a problem and not the code it was checking.  So I didn't think it was a valid failure.  Without really digging, I think you are good.

Dli3:
Looks like the eslint workflow was failing due to the change made in https://github.com/adobecom/nala/pull/488 . I've updated the version and it's working now. 
cc @amauch-adobe @skumar09

PR #492, opened 2025-03-26
fix security concers

JackySun9:
https://jira.corp.adobe.com/browse/MWPW-152026
https://jira.corp.adobe.com/browse/MWPW-152032


amauch-adobe (approved):
Approved.

Dli3 (approved):

PR #493, opened 2025-03-28
make screenshots diff to next level

JackySun9:
1. support take images as base
2. compare base vs current


runScreenDiffBase.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# Replace these variables with your actual values
+Config="visual"
```
amauch-adobe, line 4: I would recommend making these two variables as inputs from the user and then if they don't send in any, use the values you have supplied as defaults.  That way a user doesn't have to update the shell script to use it since there isn't any instructions for doing so except in the actual shell script.

runScreenDiffOne.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# Replace these variables with your actual values
+Config="visual"
```
amauch-adobe, line 4: same recommendation here, make these user inputs from the terminal and then provide the current values as defaults if nothing is sent in.

runScreenDiffBase.sh
```diff
@@ -0,0 +1,30 @@
...
+Project="chrome,ipad,iphone"
+
+# Check if category is passed as an argument
+if [ -z "$1" ]; then
+  echo "Usage: $0 <category>"
```
amauch-adobe, line 9: You might want to add some validation of user input as well to ensure it is appropriate for the call, it just helps user if they send in something wrong. Plus if it isn't valid input you can close the script here before executing the calls with bad input.

runScreenDiffOne.sh
```diff
@@ -0,0 +1,31 @@
...
+Project="chrome,ipad,iphone"
+
+# Check if category is passed as an argument
+if [ -z "$1" ]; then
+  echo "Usage: $0 <category>"
+  exit 1
```
amauch-adobe, line 10: same here you might want to add validation on user input.

tests/visual/bacom-blog/sot.test.js
```diff
@@ -27,13 +27,9 @@ test.describe('BACOM Blog SOT visual comparison test suite', () => {
         const result = await takeTwo(
           page,
           stableURL,
-          async () => {
-            await page.waitForTimeout(3000);
-          },
+          async () => { await page.waitForSelector('.feds-footer-privacyLink'); },
```
amauch-adobe, line 30: If you are waiting for an element that gets painted at the end of all sections being rendered, you can search for the hidden div at the end of all sections.  On Milo pages you would just look for the element with id="page-load-ok-milo". 

For older AEM pages there is the text in the page source "`<!-- PAGELOADOK -->`" that is rendered at the end of the HTML that can be searched for instead.  

Just a thought, if that is what you were looking for with waiting for the privacy link in the footer.

tests/visual/bacom/sot.test.js
```diff
@@ -27,13 +27,9 @@ test.describe('BACOM SOT visual comparison test suite', () => {
         const result = await takeTwo(
           page,
           stableURL,
-          async () => {
-            await page.waitForTimeout(3000);
-          },
+          async () => { await page.waitForSelector('.feds-footer-privacyLink'); },
```
amauch-adobe, line 30: same thought here as above. Not necessary, but it might be easier since those elements won't ever change.

tests/visual/cc/sot.base.test.js
```diff
@@ -0,0 +1,41 @@
...
+      for (const key of Object.keys(testdata)) {
+        const betaURL = testdata[key];
+        console.info(betaURL);
+
+        const name = `${feature.name}-${key}-${testInfo.project.name}`;
+        // eslint-disable-next-line no-await-in-loop
+        const result = await takeOne(
+          page,
+          betaURL,
+          async () => { await page.waitForSelector('.feds-footer-privacyLink'); },
```
amauch-adobe, line 31: same offered thought.

tests/visual/dc/sot.one.test.js
```diff
@@ -0,0 +1,41 @@
...
+      for (const key of Object.keys(testdata)) {
+        const betaURL = testdata[key];
+        console.info(betaURL);
+
+        const name = `${feature.name}-${key}-${testInfo.project.name}`;
+        // eslint-disable-next-line no-await-in-loop
+        const result = await takeOne(
+          page,
+          betaURL,
+          async () => { await page.waitForSelector('.feds-footer-privacyLink'); },
```
amauch-adobe, line 31: same offered thought.

tests/visual/express/sot.one.test.js
```diff
@@ -0,0 +1,38 @@
...
+
+      for (const key of Object.keys(testdata)) {
+        const betaURL = testdata[key] + MILO_LIBS;
+        console.info(betaURL);
+
+        const name = `${feature.name}-${key}-${testInfo.project.name}`;
+        const result = await takeOne(
+          page,
+          betaURL,
+          async () => { await page.waitForSelector('.feds-footer-privacyLink'); },
```
amauch-adobe, line 28: same offered thought.

amauch-adobe (approved):
Approved, but check my feedback.

runScreenDiffOne.sh
JackySun9: fixed

runScreenDiffBase.sh
JackySun9: fixed

tests/visual/bacom-blog/sot.test.js
JackySun9: I will keep this for now, we don't have pageobject for visual tests

tests/visual/cc/sot.base.test.js
JackySun9: I will keep this for now, we don't have pageobject for visual tests

tests/visual/bacom/sot.test.js
JackySun9: I will keep this for now, we don't have pageobject for visual tests

PR #532, opened 2025-09-10
add automation for brand concierge uar

JackySun9:
1. add e2e tests
2. add screenshot tests
3. add monitor test


amauch-adobe (approved):
Approved, I didn't see anything from my standpoint that needed to be updated, that being said there were a couple of lines commented out, might have been intentional. If it isn't needed it should probably be removed to keep code clean.

